PR #6, opened 2016-11-13
Add stack depth computation

MatthieuDartiailh:
Add stack depth computation based on byteplay implementation. The implementation take some margin with the stack depth (depth is always increased by 6 at the end). This comes from the byteplay implementation. @serprex may know more about this.
I had to modify some tests because the bytecode could not allow computation of the stack depth. I had no way to run the tests on Python 3.6 so the value of the wordcode should be specially checked.

### DIFF
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -495,7 +495,7 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
 
                 if last_popped_block == _BlockType.WITH_BLOCK:
                     next_stack = next_stack[:-1] + (next_stack[-1] - 1,)
-                op += [_StackState(self,  jump_targets[id(o)], next_stack,
+                op += [_StackState(self, jump_targets[id(o)], next_stack,
                                    next_block_stack, log, logging)]
 
             elif o_name == 'SETUP_LOOP':
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -338,12 +338,13 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         maxsize = 0
 
         # Analysis to carry out to determine the stack depth
-        op = [_StackState(self, logging=logging)]
+        ops = [_StackState(self, logging=logging)]
+        add_op = ops.append
 
-        while op:
+        while ops:
             # Get the nex analysis to run and update the stack depth if
             # necessary
-            cur_state = op.pop()
+            cur_state = ops.pop()
             o = sum(cur_state.stack)
             if o > maxsize:
                 maxsize = o
@@ -414,9 +415,9 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
             # Those can be SetLineno or None because we are looking at a
             # meaningless address, simply go to next opcode
             if not isinstance(o, Instr):
-                op += [_StackState(self, next_addr, cur_state.stack,
+                add_op(_StackState(self, next_addr, cur_state.stack,
                                    cur_state.block_stack, cur_state.log,
-                                   logging)]
+                                   logging))
                 continue
 
             o_name = o.name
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
I have not removed the _stacksize attribute. What should I do of this attribute ? Should stacksize be a property ?

serprex:
Note there's an important property in byteplay: the stack size only needs to be consistent for backwards jumps. Forward jumps can take the max stack size. This is important for some specialized bytecode outputs where otherwise unnecessary POPs need to be emitted. In byteplay I only implemented this for blocks of instructions before RETURN_VALUE

The _stacksize could be used if inconsistent bytecode is found. Default to None & throw if is None. Otherwise bytecode can't be used for something like [funge.py](https://github.com/serprex/Befunge/blob/master/funge.py)

PS you misspelt 'boytecode' in comment

MatthieuDartiailh:
I switched to tuples or even using append for speed and added some more comments.

Could you expand a bit more on the funge.py issue, why would bytecode fail to compute the stack size ? I would like @haypo opinion about using _stacksize, as it probably should not be used if the byte code was obtained from real code and edited before being re-converted.

@serprex could comment briefly on the extra safety at the end of adding 6 systematically to the computed stack size ? Is this mandatory or something that should be fixed ?

serprex:
It should probably be fixed, I myself am not responsible for that particular constant. [CPython](https://hg.python.org/cpython/file/tip/Python/compile.c#l4883) doesn't seem to do it. Honestly byteplay's stack depth calculation code probably needs to be rewritten

funge.py compiles befunge code into python bytecode. It uses the python stack as the befunge stack. It includes instructions to avoid stack underflow as in Python that's undefined (as in, C level undefined) & in Befunge an empty stack pops as 0. The issue is that this means there's arbitrary stack usage, since that's how Befunge works. So I just pass in a stack size of 65536

haypo:
Remarks:
* It seems like your code works on a control flow graph without using explicitly a CFG. Would it make sense to reuse ControlFlowGraph object to not have to detect basic blocks?
* Please write unit tests
* You should document your enhancement somewhere in doc/
* Would it make sense to have a stack effect method on Instr?
* Would it make sense to make compute_stacksize() public? Just document that it's slow :-)
* The code seems quite long. Maybe it deserves its own .py file?
* What is the logging thing? Is it just for debug purpose? Would it be useful to provide these logs to users?

I don't know what to think about the giant patch. I hesitate between requesting a perfect patch addressing all my comments, or accept it and then request enhancements.

At least, unit tests are required to check later that refactoring doesn't break anything.

MatthieuDartiailh:
Thanks for the feedback @haypo. 
Actually I am looking at the CPython implementation pointed out by @serprex which is much more concise and general (it appears mostly identical for Python 2 and 3). So please do not merge now. And you are indeed right in that it works on a CFG so I think porting from C may be quite straighforward using ControlFlowGraph.
I will try to re-write the stack depth computation with those ideas in mind, furthermore by having an implementation closer to CPython it will be much easier to write unit tests (currently the stack depth is often overestimated).

Other points : 
- when I am happier with the code I will add docs
- I think having stack_effect **property** (the name fits a property better) on Instr would make sense and I will add it
- we can have compute_stacksize be public (I will make it so), but what should we do of _stacksize ? What is your opinion about @serprex suggestions (fall back is compute_stacksize fail, use it if specified without recomputing) ?
- depending on my next implementation we can decide whether or not to move the code.

haypo:
> What is your opinion about @serprex suggestions (fall back is compute_stacksize fail, use it if specified without recomputing) ?

I propose to remove ConcreteBytecode._stacksize.

This attribute was a "temporary" hack before implementing compute_stacksize(). See to_code():

```
                              # FIXME: compute stack size
                              self._stacksize,
```

If someone really wants to get a specific stacksize, it's easy to call to_code() and then create a new code object to override the co_stacksize attribute.

serprex:
It isn't easy to call to_code because the code'll throw on non trivial cases

MatthieuDartiailh:
Closing in favor of #8

PR #7, opened 2016-11-13
[WIP] support for Python 2

MatthieuDartiailh:
Add support for Python 2. Changes are pretty minimal (save for the stack computation). By the way the stack computation may be wrong, I should better compare the two implementations of the stack computation in byteplay. But I would like your opinion on this first.
All tests pass locally on Python 2.

### DIFF
--- a/bytecode/__init__.py
+++ b/bytecode/__init__.py
@@ -1,9 +1,12 @@
+from __future__ import print_function, absolute_import, unicode_literals
+
 __version__ = '0.5'
 
 __all__ = ['Label', 'Instr', 'SetLineno', 'Bytecode',
            'ConcreteInstr', 'ConcreteBytecode',
            'ControlFlowGraph']
 
+import sys
 from bytecode.instr import (UNSET, Label, SetLineno, Instr, CellVar, FreeVar,   # noqa
                             Compare)
 from bytecode.bytecode import BaseBytecode, _InstrList, Bytecode   # noqa
@@ -12,11 +15,11 @@ from bytecode.concrete import (ConcreteInstr, ConcreteBytecode,   # noqa
                                _ConvertBytecodeToConcrete)
 from bytecode.cfg import BasicBlock, ControlFlowGraph   # noqa
 
+IS_PY2 = sys.version_info < (3,)
 
-def dump_bytecode(bytecode, *, lineno=False):
+def dump_bytecode(bytecode, lineno=False):
 
-    def format_line(index, line):
-        nonlocal cur_lineno, prev_lineno
+    def format_line(index, line, cur_lineno, prev_lineno):
         if lineno:
             if cur_lineno != prev_lineno:
                 line = 'L.% 3s % 3s: %s' % (cur_lineno, index, line)
@@ -25,7 +28,7 @@ def dump_bytecode(bytecode, *, lineno=False):
                 line = '      % 3s: %s' % (index, line)
         else:
             line = line
-        return line
+        return line, cur_lineno, prev_lineno
 
     def format_instr(instr, labels=None):
         text = instr.name
@@ -60,7 +63,9 @@ def dump_bytecode(bytecode, *, lineno=False):
             if lineno:
                 fields.append(format_instr(instr))
                 line = ''.join(fields)
-                line = format_line(offset, line)
+                line, cur_lineno, prev_lineno = format_line(offset, line,
+                                                            cur_lineno,
+                                                            prev_lineno)
             else:
                 fields.append("% 3s    %s" % (offset, format_instr(instr)))
                 line = ''.join(fields)
@@ -83,7 +88,10 @@ def dump_bytecode(bytecode, *, lineno=False):
                 if instr.lineno is not None:
                     cur_lineno = instr.lineno
                 line = format_instr(instr, labels)
-                line = indent + format_line(index, line)
+                line, cur_lineno, prev_lineno = format_line(index, line,
+                                                            cur_lineno,
+                                                            prev_lineno)
+                line = indent + line
             print(line)
         print()
     elif isinstance(bytecode, ControlFlowGraph):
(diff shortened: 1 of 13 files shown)


haypo:
Can you please rewrite your pull request to remove the stack computation? Only keep changes related to Python 2. And try to write a single commit.

MatthieuDartiailh:
Will do.

MatthieuDartiailh:
This is not complete yet as some tests are still broken (but mostly because they were not written with Python 2 in mind).
The main change concern the imports, the use of super,  the declaration of object as base class, the impossibility to use kwonlyargs, change to types.CodeType and the addition of enum34 as a dependency for Python2.
Before going further with refactoring the tests, I would like to know if the other changes are acceptable to you.
I will keep the PR to a single commit as requested (but you could use the squash on merge feature of Github)

bytecode/__init__.py
```diff
@@ -1,22 +1,23 @@
...
 __all__ = ['Label', 'Instr', 'SetLineno', 'Bytecode',
            'ConcreteInstr', 'ConcreteBytecode',
            'ControlFlowGraph']
 
-from bytecode.instr import (UNSET, Label, SetLineno, Instr, CellVar, FreeVar,   # noqa
-                            Compare)
-from bytecode.bytecode import BaseBytecode, _InstrList, Bytecode   # noqa
-from bytecode.concrete import (ConcreteInstr, ConcreteBytecode,   # noqa
+from .instr import (UNSET, Label, SetLineno, Instr, CellVar, FreeVar,   # noqa
+                    Compare)
```
haypo: I dislike such change. Can you try with "from __future__ import absolute_import"?

bytecode/bytecode.py
```diff
@@ -1,9 +1,7 @@
-# alias to keep the 'bytecode' variable free
-import bytecode as _bytecode
```
haypo: Ditto: try "from __future__ import absolute_import", this change is ugly :-)

bytecode/concrete.py
```diff
@@ -283,22 +284,39 @@ def to_code(self):
...
-                              tuple(self.names),
-                              tuple(self.varnames),
-                              self.filename,
-                              self.name,
-                              self.first_lineno,
-                              lnotab,
-                              tuple(self.freevars),
-                              tuple(self.cellvars))
+        if sys.version_info >= (3,):
+            return types.CodeType(self.argcount,
```
haypo: I would suggest to use an "args" list to not have to duplicate ~15 lines of code, and then use CodeType(*args).

bytecode/tests/__init__.py
```diff
@@ -113,7 +113,7 @@ def dump_bytecode(code, lineno=False):
             print()
 
 
-def get_code(source, *, filename="<string>", function=False):
```
haypo: This is one good reason to not support Python 2: it forces to write crappy API :-)

bytecode/tests/test_cfg.py
```diff
@@ -59,19 +60,32 @@ def test_constructor(self):
         self.assertBlocksEqual(code, [])
 
     def test_attr(self):
-        source = """
+        is_py2 = sys.version_info < (3,)
```
haypo: You may use the six module to get six.PY2.

But at the same time, I would prefer to avoid depending on a third party module just for one constant...

setup.py
```diff
@@ -61,6 +62,7 @@ def main():
         'author_email': 'victor.stinner@gmail.com',
         'classifiers': CLASSIFIERS,
         'packages': ['bytecode', 'bytecode.tests'],
+        'install_requires': ['enum34'] if sys.version_info < (3, 4) else []
```
haypo: This is a pain if we want to provide wheel packages. Environment markers would be better, but they require setuptools>=18.5 and pip>=6, which is another pain (not sure about setuptools version).

MatthieuDartiailh:
I managed to revert all imports to their previous form (sorry for the inconvenience). Some tests are still broken. I will try to fix them over the week-end. 
In some cases I have trouble figuring out what is the issue as even dis fail to analyse the bytecode you feed the system. This is notably true for the extended args tests.

haypo:
I started a poll on Twitter for the pull request :-)
https://twitter.com/VictorStinner/status/798469107339431936

Most people are against adding Python 2 support.

MatthieuDartiailh:
It is up to you. Let me know what you decide. I will maintain my own fork if I have to even if it is not ideal for me.

haypo:
I reject your pull request.

After 3-4 days, my Twitter poll is explicit: no Python 2 support. 80% of voters are for Python 3 only:
https://twitter.com/VictorStinner/status/798469107339431936

Two developers suggested on Twitter that you might maintain your own fork. In the past, I maintained my fork of asyncio for Python 2: Trollius. It was very different to solve every single Mercurial (and then Git) conflict because all these tiny changes required by Python 2. My suggestion is to simply forget about Python 2 and move on to Python 3. If you have to support Python 2, use a different module like codetransformer or byteplay.

I don't need Python 2 support myself and I really like Python 3 features. I want to be able to use new Python 3 features and syntax in my project.

I dislike the patch for different reasons:

* setup.py uses conditional dependency. In my experiene, it is the source of many headaches. If something goes wrong, enum34 might be installed on Python 3 which is bad. Environment markers are not supported in old versions of setuptools/pip.
* I dislike having to write "from __future__ import ..." this thing comes from the past, not the future :-)
* It's annoying to have to reimplement stdlib features like redirect_stdout()
* Having to spread "if PY3: ... else ..." all around the code is annoying too
* bytecode is very close to CPython internals, to bytecode and code objects. Supporting more Python requirements require more "if" and run tests on more Python versions, so tests take longer.
* Python 2 doesn't support keyword-only parameters, so your patch drops the "*, " from "def disassemble(source, *, filename="<string>", function=False)" whereas it was deliberate to use keyword-only parameters here to avoid mistakes, to have a saner API
* It's also a pain to have to inherit from object.
* About replacing "self.const_stack.clear()" with "del self.const_stack[:]"... Come on, the Python 2 syntax is ugly :-(
* "super().__init__()" replaced with "super(ControlFlowGraph, self).__init__()": another super annoying Python 2 requirement when you are used to Python 3 ...
* Python 2 requires to change imports, I prefer the current code

MatthieuDartiailh:
Ok this makes sense. I wish I could just drop Python 2.

Issue #9, opened 2016-11-24
Provide a way to convert frame.f_lasti into an instruction


eric-wieser:
Here's the situation - you want to find the bytecode that was just executed in a frame, possibly in a trace func. Here are a pair of things you could want:

*   The current `ConcreteInstr` object corresponding to `f_lasti`:

    ```python
    def get_concrete_index(concrete_bc, code_index):
        at = 0
        concrete_index = 0
        for c in concrete_bc:
            at += c.size
            if at < code_index:
                concrete_index += 1
        return concrete_index
    ```

    Which can be used as:
    ```python
    concrete_bc = ConcreteBytecode.from_code(frame.code)
    ci = get_concrete_index(concrete_bc, frame.f_lasti)
    concrete_instr = concrete_bc[ci]
    ```

* the current `Instr` object corresponding to `ci` or `flasti`:

    ```python
    def promote_concrete_index(bc, concrete_index):
        index = None
        at = 0
        for i, b in enumerate(bc):
            if at == concrete_index:
                index = i
            if isinstance(b, bytecode.instr.BaseInstr):
                at += 1
        return index
    ```
    Used as
    ```
    bc = concrete_bc.to_bytecode()
    i = promote_concrete_index(bc, ci)
    instr = bc[concrete_bc]
    ```

What should these functions be named, and where do they best belong as members?

eric-wieser:
Context: writing [`withhacks.frame_utils.extract_code`](https://github.com/rfk/withhacks/blob/b715b7aefcce929cf2a31d5c557b81ecfdf3375c/withhacks/frameutils.py#L87-L128)

haypo:
Hi, bytecode provides two very different kind of instructions: abstract instructions (Instr) and concrete instructions (ConcreteInstr). Which one do you want?

Building a bytecode objects from the same code is trivial, as you showed: ConcreteBytecode.from_code(frame.code)

The problem is for abstract bytecode: an abstract instruction is strongly linked to an abstract bytecode object, for example jumps uses label objects stored in the bytecode object. frame.f_lasti is hard to compute from an abstract bytecode, because you have to assemble the abstract bytecode to concrete bytecode, resolve jumps, etc.

So maybe we should limit ourself to concrete bytecode. I suggest that you add a method to ConcreteBytecode to get an instruction by its offset: return None if the offset is not exactly the start of an instruction, raise an IndexError if the offset is negative or out of the code.

Why not returning directly the instruction in your promote_concrete_index() function?

[...]

PR #10, opened 2016-11-25
Add ConcreteBytecode.{instr_at_code_offset, index_at_code_offset}

eric-wieser:
As discussed in #9.

Suggestions welcome for better method names


eric-wieser:
Re: naming:
```

index_at_code_offset(offset)?
get_instr_index(offset)
index(value=None, code_offset=None), which overloads Sequence.index
```

eric-wieser:
Test failure is `setup.py:67:1: E305 expected 2 blank lines after class or function definition, found 1`, which seems to be a problem in `master`

haypo:
Hum, I expect new user requests in the future for new methods using byte offset rather than index in the list of instructions.

I suggest to experiment a different design: add a method to get a read-only view of a ConcreteBytecode object which only works with offset. Something like:

```
view = code.offset_view()  # concrete bytecode using Python 3.6 wordcode
second_instr = view[2]  # get the ConcreteInstr at offset 2
for instr in view[10:]: print(instr)  # get instructions since the offset 10
```

Your frame.f_lasti usecase becomes obvious with such view.

It's a view, so the view is modified if the code is modified.

eric-wieser:
That's a neat idea. How would `offset_view()` work with `SetLineno` though?

haypo:
Sorry, but I still don't understand your use case :-( And I think that it's important to understand it to design the "right" API.

Can you give an example of the final expected output? Do you want to display bytecode as text? Abstract or concrete bytecode?

eric-wieser:
I want to obtain abstract bytecode corresponding to the code executed between two `frame.f_lasti`s (if possible - obviously if there are jumps leaving this segment, then erroring is fine)

PR #11, opened 2016-11-25
Add an instructions argument to ConcreteBytecode, and validate its value

eric-wieser:
Also adds validation to `Bytecode`

### DIFF
--- a/bytecode/bytecode.py
+++ b/bytecode/bytecode.py
@@ -97,7 +97,7 @@ class _InstrList(list):
 
 class Bytecode(_InstrList, BaseBytecode):
 
-    def __init__(self, instructions=[]):
+    def __init__(self, instructions=()):
         BaseBytecode.__init__(self)
         self.argnames = []
         for instr in instructions:
--- a/bytecode/bytecode.py
+++ b/bytecode/bytecode.py
@@ -97,23 +97,26 @@ class _InstrList(list):
 
 class Bytecode(_InstrList, BaseBytecode):
 
-    def __init__(self, instructions=None):
+    def __init__(self, instructions=[]):
         BaseBytecode.__init__(self)
-        if instructions is not None:
-            self.extend(instructions)
         self.argnames = []
+        for instr in instructions:
+            self._check_item(instr)
+        self.extend(instructions)
 
     def __iter__(self):
         instructions = super().__iter__()
         for instr in instructions:
-            if not isinstance(instr, (Label, SetLineno, Instr,
-                                      _bytecode.ConcreteInstr)):
-                raise ValueError("Bytecode must only contain Label, "
-                                 "SetLineno, Instr and ConcreteInstr objects, "
-                                 "but %s was found"
-                                 % instr.__class__.__name__)
-
+            self._check_item(instr)
             yield instr
+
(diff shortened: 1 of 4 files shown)


haypo:
I dislke the current code to validate instructions. The validation is delayed when you iterate on instructions. I would prefer to raise an error on .append(123) for example.

An option would be to modify InstrList to inherit from abc.MutableSequence instead of list, and implement the ABC using a private list. __setitem__() and insert() would validate that added items are instructions.

bytecode/bytecode.py
```diff
@@ -97,24 +97,27 @@ def __eq__(self, other):
 
 class Bytecode(_InstrList, BaseBytecode):
 
-    def __init__(self, instructions=None):
+    def __init__(self, instructions=[]):
```
haypo: I prefer immutable default value, so () in your case.

bytecode/bytecode.py
```diff
@@ -97,24 +97,27 @@ def __eq__(self, other):
...
-                                      _bytecode.ConcreteInstr)):
-                raise ValueError("Bytecode must only contain Label, "
-                                 "SetLineno, Instr and ConcreteInstr objects, "
-                                 "but %s was found"
-                                 % instr.__class__.__name__)
-
+            self._check_item(instr)
             yield instr
 
+    def _check_item(self, instr):
```
haypo: I suggest to rename the method to _check_instr().

bytecode/concrete.py
```diff
@@ -116,23 +116,28 @@ def disassemble(cls, lineno, code, offset):
 
 class ConcreteBytecode(_bytecode.BaseBytecode, list):
 
-    def __init__(self):
+    def __init__(self, instructions=[]):
```
haypo: Oh, it's a deliberate choice to not accept instructions here. It's because instructions depends on other Bytecode attributes: consts, names, varnames. For example, I prefer to first declare constants and then add a LOAD_CONST using a constant.

If you really want to accept a list of instructions, I suggest to accept also optional consts, names and varnames parameters.

eric-wieser: Presumably you mean other `ConcreteBytecode` attributes. You make a good point there though, perhaps this PR should be scrapped.

haypo: > perhaps this PR should be scrapped.

No, I didn't say that :-) I like your check method, I planned to do something similar.

bytecode/bytecode.py
eric-wieser: Depends if it means "check it's an instruction" or "check the item". Not too fussed either way

eric-wieser:
Ok, updated with those changes.

bytecode/bytecode.py
matrixise: Hi @eric-wieser could you use a tuple or None, and check if the value is None or an iterable and use an iterable in ``__init__`` ?

eric-wieser:
Aside from the `abc.MutableSequence` change, which would probably do better as a new PR building on this, is this good to merge?

eric-wieser:
Aside from the `abc.MutableSequence` change, which would probably do better as a new PR building on this, is this good to merge?

PR #12, opened 2016-11-25
Enable code coverage

eric-wieser:
Which is sent to https://codecov.io/gh, where some sign up might be needed

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,11 +6,11 @@ bytecode
    :alt: Latest release on the Python Cheeseshop (PyPI)
    :target: https://pypi.python.org/pypi/bytecode
 
-.. image:: https://travis-ci.org/haypo/bytecode.svg?branch=master
+.. image:: https://img.shields.io/travis/haypo/bytecode/master.svg
    :alt: Build status of bytecode on Travis CI
    :target: https://travis-ci.org/haypo/bytecode
    
-.. image:: https://codecov.io/github/haypo/bytecode/coverage.svg?branch=master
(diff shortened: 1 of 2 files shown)


eric-wieser:
Result: https://codecov.io/gh/haypo/bytecode/tree/0608d6fdb329d91f1bba0320e98df392b7e34e8a

MatthieuDartiailh:
Out of curiosity what are the advantages of using shields.io ?

eric-wieser:
Font size, badge color, and padding is identical across all badges

MatthieuDartiailh:
This looks good to me and is useful to have. May I merge this @haypo ?

MatthieuDartiailh:
It may be better to ignore coverage of tests folder ? what do you think @eric-wieser ?

eric-wieser:
Some comments on that are [here](https://softwareengineering.stackexchange.com/questions/309015/when-does-it-make-sense-to-include-test-code-in-coverage).

I'd argue this is better as it is, since it highlights which bits of the test code are not actually being run (like `WORDCODE` and most of `tests/__init__.py`)

haypo:
I dislike code coverage. I don't see how it is useful to a project.

But I'm not opposed to the change, it's up to you.

MatthieuDartiailh:
Code coverage is not a panacea but it can help identify untested part of the code. 100% coverage does not mean everything is perfect but it is better than a low coverage. For the project as it is now it does not bring much but (as the misses are not terrible but some could be fixed for example in concrete  _assemble_lnotab does not test the handling of large doff and dlineno) but it is an help to the review process of pull request.
For the above reason and since you are not opposed I am going to merge.

PR #18, opened 2017-01-05
tox : tests on 3.5 and 3.6

MatthieuDartiailh:


MatthieuDartiailh:
@haypo Not being very familiar with tox I am not sure I am doing the right thing here. I force python35 rather than python3 as I want to be sure we test on both version of the bytecode but if somebody runs on 3.4 testing 3.4 and 3.6 would be enough. Do you know a way to achieve this ?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py3, pep8
+envlist = py35, py36, pep8
```
haypo: You should remove basepython = python3 from [testenv] and add the following section:

[testenv:python3]
basepython = python3

Please also add "basepython = python3" to testenv:pep8.

tox.ini
```diff
@@ -1,5 +1,5 @@
...
```
haypo: Please keep py3, I like "tox -e py3" to use system "python3" instead of a specific Python 3 version.

tox.ini
```diff
@@ -1,8 +1,7 @@
 [tox]
-envlist = py3, pep8
+envlist = py3, py35, py36, pep8
```
haypo: Please see my previous comments, you have to define basepython=python3 for py3 and pep8 test envs.

MatthieuDartiailh:
Given that py3 is a default environment can't we omit it ?

tox.ini
MatthieuDartiailh: Given that py3 is a default environment can't we omit it ?

haypo: > Given that py3 is a default environment can't we omit it ?

Oh, you are right, it seems like "py3" is understood by tox by default. I wasn't aware of them.

Ok, in this case only pep8 should force basepython=python3.

MatthieuDartiailh: Is this then good to go ?

haypo:
Sorry, the PR now LGTM. I merged it.

MatthieuDartiailh:
Thanks !

PR #20, opened 2017-02-02
[WIP] Add a higher level handling of flags

MatthieuDartiailh:
This is a work in progress on implementing #19. I am quite busy now so it may take a while to complete this I will ping when this is ready for review.

### DIFF
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -204,7 +204,7 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         bytecode = ConcreteBytecode()
         bytecode.name = code.co_name
         bytecode.filename = code.co_filename
-        bytecode.flags = _bytecode.Flags(code.co_flags)
+        bytecode.flags = code.co_flags
         bytecode.argcount = code.co_argcount
         bytecode.kwonlyargcount = code.co_kwonlyargcount
         bytecode.first_lineno = code.co_firstlineno
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -372,9 +372,9 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         bytecode._copy_attr_from(self)
 
         nargs = bytecode.argcount + bytecode.kwonlyargcount
-        if bytecode.flags['varargs']:
+        if bytecode.flags.varargs:
             nargs += 1
-        if bytecode.flags['varkeywords']:
+        if bytecode.flags.varkeywords:
             nargs += 1
         bytecode.argnames = self.varnames[:nargs]
         _set_docstring(bytecode, self.consts)
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -284,7 +284,6 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         return cfg.compute_stacksize()
 
     def to_code(self):
-        # XXX handle Flags conversion
         code_str, linenos = self._assemble_code()
         lnotab = self._assemble_lnotab(self.first_lineno, linenos)
         nlocals = len(self.varnames)
@@ -293,7 +292,7 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
                               self.kwonlyargcount,
                               nlocals,
                               stacksize,
-                              self.flags,
+                              self.flags.to_int(self),
                               code_str,
                               tuple(self.consts),
                               tuple(self.names),
@@ -306,7 +305,6 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
                               tuple(self.cellvars))
 
     def to_bytecode(self):
-        # XXX handle flags
 
         # find jump targets
         jump_targets = set()
@@ -374,10 +372,9 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         bytecode._copy_attr_from(self)
 
         nargs = bytecode.argcount + bytecode.kwonlyargcount
-        # XXX handle flags
-        if bytecode.flags & inspect.CO_VARARGS:
+        if bytecode.flags['varargs']:
             nargs += 1
-        if bytecode.flags & inspect.CO_VARKEYWORDS:
+        if bytecode.flags['varkeywords']:
             nargs += 1
         bytecode.argnames = self.varnames[:nargs]
         _set_docstring(bytecode, self.consts)
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -205,7 +205,7 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         bytecode = ConcreteBytecode()
         bytecode.name = code.co_name
         bytecode.filename = code.co_filename
-        bytecode.flags = code.co_flags
+        bytecode.flags = _bytecode.Flags(code.co_flags)
         bytecode.argcount = code.co_argcount
         bytecode.kwonlyargcount = code.co_kwonlyargcount
         bytecode.first_lineno = code.co_firstlineno
@@ -284,6 +284,7 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         return cfg.compute_stacksize()
 
     def to_code(self):
+        # XXX handle Flags conversion
         code_str, linenos = self._assemble_code()
         lnotab = self._assemble_lnotab(self.first_lineno, linenos)
         nlocals = len(self.varnames)
@@ -305,6 +306,8 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
                               tuple(self.cellvars))
 
     def to_bytecode(self):
+        # XXX handle flags
+
         # find jump targets
         jump_targets = set()
         offset = 0
@@ -371,6 +374,7 @@ class ConcreteBytecode(_bytecode.BaseBytecode, list):
         bytecode._copy_attr_from(self)
 
         nargs = bytecode.argcount + bytecode.kwonlyargcount
+        # XXX handle flags
         if bytecode.flags & inspect.CO_VARARGS:
             nargs += 1
         if bytecode.flags & inspect.CO_VARKEYWORDS:
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -1,5 +1,4 @@
 import dis
-import inspect
 import opcode as _opcode
 import struct
 import sys
--- a/bytecode/flags.py
+++ b/bytecode/flags.py
@@ -14,19 +14,21 @@ class CoFlags(IntEnum):
     furthermore would be version dependent.
 
     """
-    CO_OPTIMIZED             = 0x0001
-    CO_NEWLOCALS             = 0x0002
-    CO_VARARGS               = 0x0004
-    CO_VARKEYWORDS           = 0x0008
-    CO_NESTED                = 0x0010
-    CO_GENERATOR             = 0x0020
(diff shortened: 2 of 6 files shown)


MatthieuDartiailh:
I still need to add some docs but a first review would be welcome @haypo.

bytecode/concrete.py
```diff
@@ -205,7 +204,7 @@ def from_code(code, *, extended_arg=False):
         bytecode = ConcreteBytecode()
         bytecode.name = code.co_name
         bytecode.filename = code.co_filename
-        bytecode.flags = code.co_flags
+        bytecode.flags = _bytecode.Flags(code.co_flags)
```
haypo: Do you really need to explicitly convert to Flags? The setter does it for you, no? If not, please revert this change.

bytecode/concrete.py
```diff
@@ -292,7 +291,7 @@ def to_code(self):
                               self.kwonlyargcount,
                               nlocals,
                               stacksize,
-                              self.flags,
+                              self.flags.to_int(self),
```
haypo: Python has standard methods to cast to int: int(obj) calls obj.__int__() or obj.__index__().

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+class CoFlags(IntEnum):
+    """Possible values of the co_flags attribute of Code object.
+
+    Notes
+    -----
+    We do not rely on inspect values here as some of them are missing and
+    furthermore would be version dependent.
+
+    """
+    CO_OPTIMIZED             = 0x00001  # noqa
```
haypo: You might remove spaces rather than repeating #noqa, but it's up to you ;-)

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
+# alias to keep the 'bytecode' variable free
+from enum import IntEnum
+from collections import defaultdict
+
+import bytecode as _bytecode
+
+
+class CoFlags(IntEnum):
```
haypo: Please rename to Flag.

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+    """Possible values of the co_flags attribute of Code object.
+
+    Notes
+    -----
+    We do not rely on inspect values here as some of them are missing and
+    furthermore would be version dependent.
```
haypo: I suggest to write: "Note: ...".

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+    object passed when converting to an integer. If no bytecode is provided the
+    default values are used.
+
+    The flags whose natural values can be inferred are:
+    - optimized
+    - generator
+    - nofree
+    - coroutine
+
+    """
```
haypo: I'm not a big fan of docstrings, it requires to duplicate the doc. The documentation in doc/ is more important than docstrings for me. You must document Flag and Flags in doc/. Then it's up to you to maintain docstrings as well, but in the project, I skipped docstrings to focus on good doc in doc/.

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def __init__(self, int_or_flags=0):
```
haypo: I suggest to rename the parameter to "value".

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+
+        self._forced = {}
+
+        if isinstance(int_or_flags, Flags):
+            self._defaults = int_or_flags._defaults.copy()
+            self._forced = int_or_flags._forced.copy()
+        elif isinstance(int_or_flags, int):
+            self._defaults = {name.split('_', 1)[1].lower():
+                              bool(int_or_flags & val)
+                              for name, val in CoFlags.__members__.items()}
```
haypo: I suggest to move "forced = {}" here.

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+        else:
+            msg = ("Flags object should be passed either None, an int  or a "
+                   "Flags instance at init, got {}.")
+            raise TypeError(msg.format(int_or_flags))
+
+    def __eq__(self, other):
+        if isinstance(other, int):
+            return self.to_int() == other
```
haypo: Hum, I would prefer to cast int to Flags, rather than the opposite. Would it work?

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+    - optimized
+    - generator
+    - nofree
+    - coroutine
+
+    """
+
+    def __init__(self, int_or_flags=0):
+
+        self._forced = {}
```
haypo: I don't understand the purpose of forced. Can you add a short comment here explaining its purpose?

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+    """Possible values of the co_flags attribute of Code object.
+
+    Notes
+    -----
+    We do not rely on inspect values here as some of them are missing and
+    furthermore would be version dependent.
+
+    """
+    CO_OPTIMIZED             = 0x00001  # noqa
+    CO_NEWLOCALS             = 0x00002  # noqa
```
haypo: I suggest to remove CO_ prefix.

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+    CO_VARARGS               = 0x00004  # noqa
```
haypo: I dislike hardcoded values. I would prefer to get them from a Python module, like dis.COMPILER_FLAG_NAMES, if possible.

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+    CO_VARKEYWORDS           = 0x00008  # noqa
+    CO_NESTED                = 0x00010  # noqa
+    CO_GENERATOR             = 0x00020  # noqa
+    CO_NOFREE                = 0x00040  # noqa
+    # New in Python 3.5
+    CO_COROUTINE             = 0x00080  # noqa
+    CO_ITERABLE_COROUTINE    = 0x00100  # noqa
```
haypo: Maybe only define them if the running Python version if >= 3.5? I don't know if bytecode has a policy to support a bytecode generated by a different version than the running Python.

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+        """The compiled code block has a varargs argument.
+
+        """
+        return self._forced.get('varargs', self._defaults['varargs'])
+
+    @varargs.setter
+    def varargs(self, value):
+        if not isinstance(value, bool):
+            raise ValueError('Flag value should be a boolean')
+        self._forced['varargs'] = value
```
haypo: Maybe you can factorize these methods with __getattr__() and __setattr__() methods, the code looks to be duplicated for each flag.

bytecode/tests/test_flags.py
```diff
@@ -0,0 +1,103 @@
...
+
+class FlagsTests(unittest.TestCase):
+
+    def test_instantiation(self):
+
+        empty_flag = Flags()
+        self.assertFalse(any(empty_flag._defaults.values()))
+        self.assertFalse(empty_flag._forced)
+
+        flags = Flags(CoFlags.CO_OPTIMIZED + CoFlags.CO_NESTED)
```
haypo: Hum, for flags, I prefer to write "a | b" than "a + b".

bytecode/tests/test_flags.py
```diff
@@ -0,0 +1,103 @@
...
+        for m in CoFlags.__members__:
+            expected = (True
+                        if m in {'CO_OPTIMIZED', 'CO_NESTED'} else
+                        False)
+            self.assertEqual(flags.get_default(m.split('_', 1)[1].lower()),
```
haypo: I would prefer a class method to get "all flags" than duplicating this split/lower code.

bytecode/tests/test_flags.py
```diff
@@ -0,0 +1,103 @@
...
+        for val in CoFlags.__members__.values():
+            self.assertEqual(Flags(val).to_int(), val)
+
+        with self.assertRaises(ValueError):
+            Flags().to_int(Bytecode())
+
+        # Infer optimized and nofree
+        forced_flag = Flags()
+        forced_flag.optimized = False
+        forced_flag.nofree = False
```
haypo: Please instanciate ConcreteBytecode here:

bytecode = ConcreteBytecode()

bytecode/flags.py
```diff
@@ -0,0 +1,297 @@
...
+    CO_COROUTINE             = 0x00080  # noqa
+    CO_ITERABLE_COROUTINE    = 0x00100  # noqa
+    # New in Python 3.6
+    CO_ASYNC_GENERATOR       = 0x00200  # noqa
+
+    # __future__ flags
+    CO_FUTURE_GENERATOR_STOP = 0x80000  # noqa
+
+
+_CAN_DEDUCE_FROM_CODE = ('optimized', 'generator', 'nofree', 'coroutine')
```
haypo: "Can"? It seems like if you pass a concrete bytecode object, these flags are *always* set, no?

I'm asking if we cannot find a better name for the constant.

haypo:
IMHO once we agree on the Flags design, maybe bytecode API would be complete enough for a version 1.0? What do you thing?

By the way, TODO.rst seems oudated: "ConcreteBytecode.to_code(): compute the stack level, see byteplay and Python/compile.c?" this item is now implemented, no?

Flags and stack depth were the two remaining blocker point to stabilize the API.

I'm not sure that bytecode has enough users to stabilize the API, but maybe it would have more users with a stable API? :-)

bytecode/concrete.py
MatthieuDartiailh: will do

bytecode/concrete.py
MatthieuDartiailh: The point here was to pass the bytecode to infer some flag values. If we want to rely on standard methods we need to set the bytecode as an attribute of the flag object.

bytecode/flags.py
MatthieuDartiailh: will do

bytecode/flags.py
MatthieuDartiailh: The point here is that the proper value of the flag can be deduced from the associated bytecode, so it is really a 'can'. 'Generator' and 'coroutine' are a bit special as they can be ambiguous (cf async_generator and iterable_coroutine)

bytecode/flags.py
MatthieuDartiailh: I will add the docs in doc but I like docstrings so if it is fine with you I will leave them.

bytecode/flags.py
MatthieuDartiailh: forced is used to keep track of flags whose values have been set by the user and whose value should hence not be inferred from the bytecode.

bytecode/flags.py
MatthieuDartiailh: The issue is that if one value has been 'forcibly' set it will never match the int.

bytecode/flags.py
MatthieuDartiailh: Will try to come up with something a bit less redundant yes.

bytecode/tests/test_flags.py
MatthieuDartiailh: make sense

bytecode/tests/test_flags.py
MatthieuDartiailh: Done

MatthieuDartiailh:
This is superseded by #23, so closing

PR #23, opened 2017-05-08
Add support for flag handling using IntFlag from aenum

MatthieuDartiailh:
Following @serhiy-storchaka suggestion I tried to move the flag handling over to IntFlag. The PR is not yet polished but the basics are here. Some points on which I would like to have @haypo feelings:
- IntFlag is new in 3.6 so I had to backport it for lower version, is that ok with you ?
- a bytecode object can now update its flag, but it does not update any inner code object. Does that make sense or should the update be recursive ?
- byteplay was doing some quite crazy things to infer the value of NESTED (cf https://github.com/serprex/byteplay/blob/master/wbyteplay.py#L584) do we want something similar ?

I am sorry it took me so long to do this, I am writing my thesis manuscript so I am a bit busy lately but I will try to finish this as soon as get some feedback.


MatthieuDartiailh:
Another option would be to add a dependency on aenum which provides IntFlag for all version of python and  stdlib compatible enums. @haypo what is your opinion about dependencies ? If I remember correctly you were not a big fan of per version dependency, so do you prefer aenum always or the bundled backport ?

haypo:
I wrote the bytecode project a few months ago, but I'm not using actually. I have no strong opinion on the best implementation. It's up to you.

MatthieuDartiailh:
Thanks. I will go with aenum and let you know when everything is properly polished with documentation and all.

MatthieuDartiailh:
I finally took some time to add the missing documentation. From my end this is good to go. I f you could give just a quick look @haypo that would be great. I documented the enumeration as a class as I could not find any other way to do so.

MatthieuDartiailh:
@haypo do you want a look or should I merge as is ?

MatthieuDartiailh:
@haypo I will consider your long silence as a yes and merge this in a week.

haypo:
In short, I'm not interested to work on the bytecode project anymore, so feel free to modify it whatever you want :-) I don't care much how flags are passed.

MatthieuDartiailh:
Ok thanks

PR #26, opened 2018-03-31
remove useless EXTENDED_ARG 0

SnoopyLane:
to_bytecode() now removes EXTENDED_ARG 0 instructions that
appear in the input.  These are useless, but can be left behind
by ConcreteBytecode.from_code.  Removing these from the input
prevents analyzers from having to deal with EXTENDED_ARG
instructions.  Most of them are already removed, and having
extras laying around causes failure when converting back
to code.


MatthieuDartiailh:
I do not believe this is the right fix see #28

Issue #32, opened 2018-07-20
Support general constants


thautwarm:
Hello, 
I've found that the code object could not take unhashable object as its co_consts, for you use a dictionary to store the constants.
I think that we may not follow the implementation of `_PyCode_ConstantKey` for further usage. Actually my project needs making more kinds of constants than CPython already has.
https://github.com/vstinner/bytecode/blob/a7cc7a52ca10e58a8aee48052edd069fac3c5a01/bytecode/concrete.py#L439

```python
    self.consts = {}
    ...
    def add_const(self, value):
        key = const_key(value)
        if key in self.consts:
            return self.consts[key]
        index = len(self.consts)
        self.consts[key] = index
        return index
```

Could we add a compiler flag to support unhashable constants? 
Just consider using associate list (`List[Tuple[K, V]]`)  to store unhashable ones.

MatthieuDartiailh:
The issue here is not so much unhashable constants but more non-constant constants. I don't really like this. Could you detail your use case for what looks like an abuse of the constants to me ?

thautwarm:
@MatthieuDartiailh Sure. 

Actually we're making a transpiler from an extension of SQL to python, which requires loading tables when applying nested table join (like `SELECT xxx from a inner join b on cond(a,b) left outer join c on cond(ab, c)`). We want to avoid the overhead of function call so we use nested for-loop to implement it, like
```python
for record_tb1 in tb1:
   for record_tb2 in tb2:
      ...
```
For `tb1` and `tb2` is actually constant, we don't want to load it from global to slow down program heavily.

AFAICS, there're at least following three reasons why we should supply the way to extend python constants:
- As python doesn't have static constants like C++, loading a variable from global scope just costs unnecessarily. 
- A global readonly variable of any type is actually a constant. For example, In the whole program there could be a singleton class which is permanent and it's just okay. Now python just has literal constants, and it hurts sometimes.
- A constant is decided when its context is defined, but the global variable could be rewrited, which leads to the danger of visiting unexpected variable. The global scoping is just wrong for the existence of perspective mutation. 

People may want to use a singleton class with readonly properties to imitate constants, however the overhead just bursts, while it's verbose and the name binding of the class could be rewrited, too.

MatthieuDartiailh:
Could you clarify one point ? How do you use the bytecode ? Reading your previous message I would guess you need to generate valid Python code at the end, no ?

thautwarm:
[...]

PR #33, opened 2018-07-20
support general constants in code object.

thautwarm:
#32

### DIFF
--- a/bytecode/concrete.py
+++ b/bytecode/concrete.py
@@ -4,8 +4,10 @@ import opcode as _opcode
 import struct
 import sys
 import types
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
As mentioned in #32 I believe we can do simpler simply by tweaking the key generation function. If you have some time to look at it, that would be great otherwise I  will come back to it but I can make no promise on when.

thautwarm:
@MatthieuDartiailh Sorry for my delay. 
Keeping a simple dictionary does make sense, and a corresponding implementation would be presented here recently.

PR #36, opened 2018-09-01
introduce a const_list to avoid the loss of readability when accessing constants from `_ConvertBytecodeToConcrete`

thautwarm:
@serhiy-storchaka 
You mean this sort of implementation? https://github.com/vstinner/bytecode/pull/34#discussion_r213190168


MatthieuDartiailh:
This is covered as part of #34 but I will add the copy I omitted since it looks cleaner.

PR #42, opened 2018-10-16
Add syntax coloration for the Python snippet

Naereen:
Add syntax coloration for the Python snippet


MatthieuDartiailh:
Thanks for the fix !

PR #43, opened 2018-10-16
Fix stack effect

thautwarm:
Fix https://github.com/vstinner/bytecode/issues/38.
@MatthieuDartiailh 
If `dis.stack_effect` doesn't convert oparg when unnecessarily, we wouldn't get this error.

https://github.com/python/cpython/blob/137b0632dccb992ca11e9445142fb33a29c33a51/Modules/_opcode.c#L38

However, for we cannot change the standard module immediately, we have to avoid passing the oparg when we didn't calculate the operand of an instruction.

For instance,  `Instr('LOAD_CONST', 0xFFFFFFFFFFFFFFF)` might donate bytecode `LOAD_CONST 1`, and when we calculate its stack effect, we should use `dis.stack_effect(100, 1)` instead of `dis.stack_effect(100, 0xFFFFFFFFFFFFFFF)`. 

Actually we cannot convert `0xFFFFFFFFFFFFFFF` to `1` when calculating stack effects, fortunately if we have to convert the oparg of `Instr` to the  operand of bytecode instruction, it means that the oparg is insignificant to the result.

Thanks @serhiy-storchaka for
https://github.com/python/cpython/blob/137b0632dccb992ca11e9445142fb33a29c33a51/Python/compile.c#L859.


SnoopyLane:
> For instance, `Instr('LOAD_CONST', 0xFFFFFFFFFFFFFFF)` might donate bytecode `LOAD_CONST 1`, and when we calculate its stack effect, we should use `dis.stack_effect(100, 1)` instead of `dis.stack_effect(100, 0xFFFFFFFFFFFFFFF)`.

This is a subtle but critical detail.  Thanks for the reminder!

bytecode/tests/test_instr.py
```diff
@@ -212,6 +212,23 @@ def test_stack_effects(self):
             if not instr.has_jump():
                 self.assertEqual(jump, no_jump, msg)
 
+        # LOAD_CONST uses a concrete python object as its oparg, however, in
+        #       dis.stack_effect(opcode.opmap['LOAD_CONST'], oparg),
+        # oparg should be the index of that python object in the constants.
+        #
+        # Fortunately, for an instruction whose oparg isn't equivalent to its
+        # form in binary files(pyc format), the stack effect could be a
```
MatthieuDartiailh:
```suggestion
        # form in binary files(pyc format), the stack effect is a
```

bytecode/tests/test_instr.py
```diff
@@ -212,6 +212,23 @@ def test_stack_effects(self):
...
+        # constant which has nothing to do with its oparg.
```
MatthieuDartiailh:
```suggestion
        # constant which does not depend on its oparg.
```

bytecode/tests/test_instr.py
```diff
@@ -212,6 +212,23 @@ def test_stack_effects(self):
...
+        #
+        # The second argument of dis.stack_effect could not be
```
MatthieuDartiailh:
```suggestion
        # The second argument of dis.stack_effect cannot be
```

bytecode/tests/test_instr.py
```diff
@@ -212,6 +212,23 @@ def test_stack_effects(self):
...
+        # more than 2**31 - 1. If stack effect of an instruction is
+        # independent of its oparg, we pass 0 as the second argument
+        # of dis.stack_effect.
+        # (As a result we can calculate stack_effect for
+        #  any LOAD_CONST instructions)
```
MatthieuDartiailh:
```suggestion
        #  any LOAD_CONST instructions, even for large integers)
```

bytecode/instr.py
```diff
@@ -306,9 +316,15 @@ def stack_effect(self, jump=None):
         # then we should pass that through, and perhaps remove some of the
         # overrides that are set up in _init_stack_effects()
 
-        # All opcodes whose arguments are not represented by integers have
-        # a stack_effect indepent of their argument.
-        arg = (self._arg if isinstance(self._arg, int) else
+        # Each of following opcodes has a stack_effect indepent of its
```
MatthieuDartiailh:
```suggestion
        # Each of the following opcodes has a stack_effect independent of its
```

bytecode/instr.py
```diff
@@ -171,6 +171,16 @@ def _check_arg_int(name, arg):
         _opcode.opmap['SETUP_FINALLY']: (6, 9),
     })
```
MatthieuDartiailh:
```suggestion
# Opcodes for which the value of the argument has an impact on the stack effect
```

bytecode/instr.py
thautwarm: I do want to accept all of your suggestions, but it seems that I have to make lots of commits  because each of suggestion application performs a commit.

Did I break PEP8? Or just my word is verbose...

MatthieuDartiailh: Do not worry about the commits I will most likely squash them when merging.

thautwarm: @MatthieuDartiailh Oh, I've just made a new commit from the local repo... My bad.

PR #47, opened 2019-08-20
Add Python 3.8 testing

MatthieuDartiailh:


laike9m:
Hi Matthieu, is bytecode ready to support Python 3.8?

MatthieuDartiailh:
Should be yes. The documentation needs to be updated a bit though.

laike9m:
Good to hear. Hope to see a new release soon cause my project is relying on bytecode 😆

MatthieuDartiailh:
Need to find some time but it is my list of things to do yes. As a user could you comment on #48 ?

laike9m:
Still waiting for a new release...

MatthieuDartiailh:
As I mentioned the docs need to be updated first and since I maintain this project (and other open source project) on my free time it may be a bit slow. If you have time to contribute a PR to the docs underlying the difference between python 3.8 bytecode and previous version that would speed things up. Otherwise it will be another week or two before I can do the edits and the release (but no more than that).

laike9m:
If you can provide some guide/pointers, I can try.

MatthieuDartiailh:
A good place to start is https://docs.python.org/3.8/whatsnew/3.8.html#cpython-bytecode-changes and from there update the documentation (examples in usage.rst need to be updated and this could actually be expanded to also describe try/except/finally structure if you feel up to it).

laike9m:
Thanks Matthieu, however I've been busy fixing some issues in my own project caused by 3.8 😢

Issue #45, opened 2018-10-25
Docs: update examples for Python 3.8


MatthieuDartiailh:
cf comment by @serhiy-storchaka 

https://github.com/vstinner/bytecode/pull/41#issuecomment-431392179

MatthieuDartiailh:
Closed by #50

PR #49, opened 2019-10-31
Update doc example

laike9m:
Fixed simple loop example for Python 3.8. #45


laike9m:
Only updated one example.

laike9m:
I think this is only place that needs to be updated.

MatthieuDartiailh:
Superseded by #50

PR #53, opened 2020-01-27
Add Compare to __all__

Wooza:
Compare is a part of the public api and should be in __all__.
Is there a reason for it not being in there?


MatthieuDartiailh:
Seems like an oversight to me. Could you add an entry to the release notes. I will merge shortly.

Issue #51, opened 2020-01-06
treating `const_key` of a code object as mutable one


thautwarm:
https://github.com/vstinner/bytecode/blob/45d643f6a706ae95586a874ea0fbcb18e797bfa1/bytecode/instr.py#L39

If the code object's constant pool contains mutable data(even `list`s), the return const key cannot get hashed.
I for one got this:
```python
  File ".../bytecode/concrete.py", line 466, in add_const
    if key in self.consts_indices:
TypeError: unhashable type: 'list'
```

thautwarm:
I suggest removing `types.CodeType` here.

thautwarm:
As the tuple and frozenset could consist of mutable data, following `return (obj_type, obj, key)` wouldn't work in the above case.

https://github.com/vstinner/bytecode/blob/45d643f6a706ae95586a874ea0fbcb18e797bfa1/bytecode/instr.py#L69

https://github.com/vstinner/bytecode/blob/45d643f6a706ae95586a874ea0fbcb18e797bfa1/bytecode/instr.py#L73

MatthieuDartiailh:
We end up having an issue with CodeType because we allow to build weird code type that end up being unhashable which usually never happens. I am not sure what relying on id for CodeType could lead too and why CPython does not do it (an idea @serhiy-storchaka ?). Depending on the answer we may need to "legalize" the CodeType to be able to use it as key.

Regarding tuple and frozenset your argument seem valid, I would need to think more about possible corner cases. Could you make a PR in the meantime ? 

For reference the current CPython implementation is here : https://github.com/python/cpython/blob/master/Objects/codeobject.c#L678

thautwarm:
>  I am not sure what relying on id for CodeType could lead too and why CPython does not do it

For CPython implementation, all constants can be serialized by `marshal.dumps`. So keeping this function simple suffices and is more efficient.

As a user of `bytecode`, IMO we're working with code objects and trying to do optimizations on them. Constant linking to avoid loading variables from global scope is really a huge performance gain in many cases, so it's indeed  a use case.

> Regarding tuple and frozenset your argument seem valid, I would need to think more about possible corner cases. Could you make a PR in the meantime ?

Yes. In fact I have come up with a proven method to implement `const_key` . I'll explain in the PR later.

PR #54, opened 2020-01-31
Simplifying and enhancing const_key

thautwarm:
Explanation
=====================================

using `marshal.dumps` for `const_key`
---------------------------------------------------------

In CPython, code objects compiled from a source code file, are able to get serialized into binary contents via `marshal.dumps`.

This gives us some useful observations:

Constants(all elements in `code.co_consts`) of every code object, are also able to get serialized into binary contents via `marshal.dumps`.

`marshal.dumps` must be able to distinguish data of different types and values, even for something we might make mistakes on, e.g., https://github.com/vstinner/bytecode/issues/32#issuecomment-413537179, and keep the property of data with structural equality. This is because, `marshal.dumps` is a foundation of CPython: CPython relies on this to keep the same behaviour of `.py` and `.pyc`.

Hence, `marshal.dumps` can precisely give us a perfect encoding that can distinguish objects from each other and keep the equality of structure data, for all constants supported by Python. On the other side, the encoding itself is a simple `bytes`, thus, if we just want to keep the
same behaviour as CPython's, `marshal.dumps` can sufficiently become an implementation of the function `const_key` mentioned in https://github.com/vstinner/bytecode/issues/51 .

```python
def const_key(const):
    return marshal.dumps(const)
# or const_key = marshal.dumps
```

**This already keeps the total compatibility to CPython**.

More general constants in bytecode, and compatibility
----------------------------------------------------------

However, for [this good reason](https://github.com/vstinner/bytecode/issues/32#issuecomment-408626651) given by @SnoopyLane , this will be very beneficial:

> Let the bytecode library support construction of valid bytecodes, even if that means the result is not valid Python.

Hence we can also support non-`marshal`-able constants in code objects.

This time, my proposal for this extension is,

```python
def const_key(const):
    try:
        marshal.dumps(const)
    except ValueError: # un-marshal-able
        return type(const), id(const)
```

`type(const), id(const)` is also the same strategy used before: https://github.com/vstinner/bytecode/blob/45d643f6a706ae95586a874ea0fbcb18e797bfa1/bytecode/instr.py#L37 .

Advantage of this implementation
-----------------------------------------------------

1. **No need any more to synchronize our code for `const_key` with `_PyCode_ConstantKey`, but keep full compatibility to CPython**.

2. A little more efficient when we're calculating keys for `tuple` or other built-in structural data.

3. Code size for `const_key` gets reduced, also, imports of modules like `math` and `types` can get eliminated.

### DIFF
--- a/bytecode/instr.py
+++ b/bytecode/instr.py
@@ -4,6 +4,7 @@ import math
 import opcode as _opcode
 import sys
 import types
+from marshal import dumps as _dumps
 
 import bytecode as _bytecode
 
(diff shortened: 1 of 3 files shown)


thautwarm:
fix https://github.com/vstinner/bytecode/issues/51

MatthieuDartiailh:
Could you please add tests covering #51 to ensure the new behavior fixes it ? And please update the changelog too.

thautwarm:
@MatthieuDartiailh Done. Sorry for the late PR.

MatthieuDartiailh:
Thanks ! I will do my best to release over the week-end.

thautwarm:
@MatthieuDartiailh Thank you, and take your time!

PR #64, opened 2020-07-18
(Slightly) improve support for EXTENDED_ARG

IPSW1:
Improves the use of EXTENDED_ARG, currently only in combination with UNPACK_EX to limit side effects.

Add tests for successful and also one failing case.


IPSW1:
Sorry, I am not sure what is going on with this failing configuration.

IPSW1:
Better solution in #65

PR #65, opened 2020-07-19
Prevent EXTENDED_ARG to propagate past ConcreteBytecode

### LINKED ISSUE #63: Stack size with EXTENDED_ARG
Hi,
I stumbled upon a problem with the stack size computation in combination with the `EXTENDED_ARG` instruction. Take for example this code snipped:
```
p = [1, 2, 3, 4, 5, 6]
q, r, *s, t = p
print(q, r, s, t)
```
```
>>> 1 2 [3, 4, 5] 6
```

The disassembly looks like this:
```
1           0 LOAD_CONST               1 (1)
            2 LOAD_CONST               2 (2)
            4 LOAD_CONST               3 (3)
            6 LOAD_CONST               4 (4)
            8 LOAD_CONST               5 (5)
            10 LOAD_CONST              6 (6)
            12 BUILD_LIST              6
            14 STORE_FAST              1 (p)
                  
2          16 LOAD_FAST                1 (p)
           18 EXTENDED_ARG             1
           20 UNPACK_EX              258
           22 STORE_FAST               2 (q)
           24 STORE_FAST               3 (r)
           26 STORE_FAST               4 (s)
           28 STORE_FAST               5 (t)
                  
3          30 LOAD_GLOBAL              0 (print)
           32 LOAD_FAST                2 (q)
           34 LOAD_FAST                3 (r)
           36 LOAD_FAST                4 (s)
           38 LOAD_FAST                5 (t)
           40 CALL_FUNCTION            4
           42 POP_TOP
           44 LOAD_CONST               0 (None)
           46 RETURN_VALUE

```

Now generating the bytecode from this code and converting back to a code object fails due to wrong stack size computation, when the bytecode is generated with the parameter to include extended arguments [here](https://github.com/vstinner/bytecode/blob/f9bc2950ac3a2185d93249dcde8649eba5fb6cec/bytecode/concrete.py#L197) set to True (I specifically need this behavior for the bytecode to match the disassembly).

The problem seems to be the use of the `EXTENDED_ARG` instruction, which is not handled correctly. The stack size computation treats `UNPACK_EX` as a single instruction and takes into account only the lower byte (corresponding to the number of variables before the list value, i.e. variables `q`, `r`), but not the higher byte provided by `EXTENDED_ARG` (corresponding to the number of values after the list value, i.e. variable `t`). This leads to a negative stack size.

Is this a known problem and is there a solution or a workaround for this, except just excluding the `EXTENDED_ARG` instruction?

(For reference: [`UNPACK_EX` documentation](https://docs.python.org/3/library/dis.html#opcode-UNPACK_EX), [stack effect of `UNPACK_EX`](https://github.com/python/cpython/blob/3b6a8d2455c6897085f4277737b0f9b9a3847c24/Python/compile.c#L999))

EDIT: Ok, an easy workaround would be to generate two bytecode objects, one without and one with the parameter for extended arguments. I can generate a code object from the one without `EXTENDED_ARG` and keep the other one for reference and comparison with the disassembly. Nevertheless, I don't think that is the nicest option here.

MatthieuDartiailh:
As discussed in https://github.com/vstinner/bytecode/issues/63, some users prefer to generate ConcreteBytecode that preserve the existence of EXTENDED_ARG for ease of comparison with dis.dis. This operation is possible by passing `extended_args=True` to `ConcreteBytecode.from_code`.  Currently this representation is preserved in Bytecode and ControlFlowGraph.

However preserving such a low-level detail in more abstract representation can lead to a lot of problem since they impact the handling of the stack size computation, long jumps, etc. In this PR, I propose to simply disallow EXTENDED_ARG outside of ConcreteBytecode, by doing the following:
- forbid EXTENDED_ARG in Instr
- forbid ConcreteInstr in Bytecode and ControlFlowGraph.
- enforce that when converting a ConcreteBytecode to a Bytecode object we remove the EXTENDED_ARG
This will lead to more consistency and will avoid to add a lot of extra logic to handle EXTENDED_ARG in abstract representation.

This is a non backward compatible change but I do not see the use case for mixing concrete and abstract instructions, so I assume the inconvenience will be minor. Any opinion @thautwarm @laike9m @serhiy-storchaka ?

Closes #63


MatthieuDartiailh:
In the absence of comments I will merge in 7 to 10 days.

thautwarm:
LGTM. I expect `ConcreteBytecode` as-is and `Bytecode`  is for high level use. Even without this PR I use `bytecode` this way.

Issue #68, opened 2020-10-12
Stack size computation issue


Z-Shang:
For example:
![image](https://user-images.githubusercontent.com/859939/95697119-66d1a400-0c89-11eb-9897-32ee2a46d58d.png)

This will cause a stack underflow, but no error was given until the segfault.
Expecting something like:
![image](https://user-images.githubusercontent.com/859939/95697310-1149c700-0c8a-11eb-94d4-e13dae235bd2.png)

MatthieuDartiailh:
bytecode does not currently perform this kind of validation (which would require to know more about each bytecode than is currently known). I am not sure I would want such validation ran by default but it could be interesting. Feel free to put together a PR, I will be happy to review it.

Z-Shang:
PR created: https://github.com/vstinner/bytecode/pull/69 :)

PR #69, opened 2020-10-15
quick and dirty way of checking pre and post stack effect

Z-Shang:
As mentioned in [this issue](https://github.com/vstinner/bytecode/issues/68),
I've modified compute stack size to take in consideration of the prerequisite stack size for some ops (e.g. `BUILD_CONST_KEY_MAP`)

For these definitely failing cases:
image

Before:
image

After:
image


MatthieuDartiailh:
This looks reasonable enough. I will give a more detailed read later and try to remember why NOP used to be a problem.

Z-Shang:
> This looks reasonable enough. I will give a more detailed read later and try to remember why NOP used to be a problem.

Thank you 👍

MatthieuDartiailh (changes requested):
Some quick nitpick I caught through this a second time. Next I need to sit and check that we do catch all opcodes (for all supported Python version). I wish CPython would provide more tooling to support this.

bytecode/instr.py
```diff
@@ -282,6 +282,27 @@ def stack_effect(self, jump=None):
         else:
             return dis.stack_effect(self._opcode, arg, jump=jump)
 
+    def pre_and_post_stack_effect(self, jump=None):
+        def _pushes_back(opname):
```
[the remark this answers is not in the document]
Z-Shang: sure, I'll take a look at how the updates in Python 3.9 affect this

bytecode/instr.py
```diff
@@ -282,6 +282,27 @@ def stack_effect(self, jump=None):
...
+                or opname.startswith("GET_")
+                or opname.startswith("BINARY_")
+                or opname.startswith("INPLACE_")
+                or opname.startswith("BUILD_")
+                or opname.startswith("CALL_")
+            )
+
+        _effect = self.stack_effect(jump=jump)
+        # To compute pre size and post size to avoid segfault cause by not enough stack element
+        if _opcode.opname[self._opcode].startswith("DUP_TOP"):
```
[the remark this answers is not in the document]
Z-Shang: by changing `pushes_back` to a free function, this will only be used here.

bytecode/instr.py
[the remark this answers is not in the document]
Z-Shang: I've added the Python 3.9 new ops, but by hard coding doesn't feel right.
We may have to find a better way in the future

bytecode/tests/test_concrete.py
```diff
@@ -1309,7 +1309,8 @@ def test_compute_jumps_convergence(self):
         code = Bytecode()
         label1 = Label()
         label2 = Label()
-        nop = "UNARY_POSITIVE"  # don't use NOP, dis.stack_effect will raise
+        nop = "NOP"  # don't use NOP, dis.stack_effect will raise
+        # Why? Changing to NOP seems to have no harm
```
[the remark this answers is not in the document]
Z-Shang: done

bytecode/instr.py
```diff
@@ -282,6 +282,37 @@ def stack_effect(self, jump=None):
         else:
             return dis.stack_effect(self._opcode, arg, jump=jump)
 
+    def pushes_back(self):
```
MatthieuDartiailh, line 285: Should have a leading underscore, this is meant to be private. And it does not need to be a method, we could extract the opcode name once, and pass it to an independent function, but this should be fine.

bytecode/instr.py
```diff
@@ -282,6 +282,37 @@ def stack_effect(self, jump=None):
...
+            "DICT_MERGE",
+            "IS_OP",
+            "CONTAINS_OP",
+        ]
+
+    def pre_and_post_stack_effect(self, jump=None):
+
+        _effect = self.stack_effect(jump=jump)
+        # To compute pre size and post size to avoid segfault cause by not enough stack element
+        if _opcode.opname[self._opcode].startswith("DUP_TOP"):
```
MatthieuDartiailh, line 310: We may want to handle the ROT_ opcodes in the same way since I assume the interpreter crashes if there are not enough objects on the stack.

bytecode/tests/test_bytecode.py
```diff
@@ -253,6 +253,34 @@ def test_to_code(self):
...
+            code.compute_stacksize()
+
+    def test_negative_size_call(self):
+        code = Bytecode()
+        code.first_lineno = 1
+        code.extend([Instr("CALL_FUNCTION", 0)])
+        with self.assertRaises(RuntimeError):
+            code.compute_stacksize()
+
+    def test_negative_size_build(self):
```
MatthieuDartiailh, line 277: I would like to see some extra tests in particular for corner cases like DUP_TOP.

MatthieuDartiailh (changes requested):

MatthieuDartiailh (changes requested):
Can you also rebase on master ? I will do my best to give this the thorough read it deserves next week.

bytecode/instr.py
```diff
@@ -34,6 +34,27 @@ def const_key(obj):
...
+        or opname.startswith("CALL_")
+    ) or opname in [
+        "LIST_TO_TUPLE",
+        "LIST_EXTEND",
+        "SET_UPDATE",
+        "DICT_UPDATE",
+        "DICT_MERGE",
+        "IS_OP",
+        "CONTAINS_OP",
+    ]
```
MatthieuDartiailh, line 55: I believe the following are missing from the list of push back (You could use a tuple rather than a list by the way):
SET_ADD
LIST_APPEND
MAP_ADD
IMPORT_
MAKE_FUNCTION
FORMAT_VALUE

bytecode/instr.py
```diff
@@ -282,6 +303,19 @@ def stack_effect(self, jump=None):
...
+
+        _effect = self.stack_effect(jump=jump)
+        # To compute pre size and post size to avoid segfault cause by not enough stack element
+        _opname = _opcode.opname[self._opcode]
+        if _opname.startswith("DUP_TOP"):
+            return _effect * -1, _effect * 2
+        if _pushes_back(_opname):
+            return _effect - 1, 1
+        return {"ROT_TWO": (-2, 2), "ROT_THREE": (-3, 3), "ROT_FOUR": (-4, 4)}.get(
+            _opname, (_effect, 0)
```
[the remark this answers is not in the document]
MatthieuDartiailh: UNPACK_* opcodes need to be handled separately since they consume one value and pushed multiple ones

MatthieuDartiailh (changes requested):
I noticed a bunch of possible issues. Please have a look and try to add additional tests to cover those.

bytecode/instr.py
Z-Shang: it is quite tricky for `IMPORT_` OPs
as `IMPORT_NAME` pops 2 and pushes 1
`IMPORT_FROM` pops 0 and pushes 1
`IMPORT_STAR` simply pops 1 and pushes nothing.

I think it then makes sense to consider only `IMPORT_NAME`

Z-Shang:
I'm just wondering, it is possible to get the upstream (id est the Python team) to support this as a part of the standard library (such as `dis.stack_effect`)...

bytecode/instr.py
MatthieuDartiailh: I agree for IMPORT_

MatthieuDartiailh:
Once we consolidate that here I will try to reach to the Python developers to suggest something along those lines indeed.

MatthieuDartiailh:
Changes look nice. I will give it another thorough read during the week and convince myself it will always work (some exception handling have been worrying me but since the tests are passing I think everything is fine and just need to go through the logic once again myself). Hopefully I will merge once I am done. Could you add an entry to the change log in the meantime ?

Z-Shang:
> Changes look nice. I will give it another thorough read during the week and convince myself it will always work (some exception handling have been worrying me but since the tests are passing I think everything is fine and just need to go through the logic once again myself). Hopefully I will merge once I am done. Could you add an entry to the change log in the meantime ?

for the changelog, do I also have to update the version number?

MatthieuDartiailh:
You can just keep adding below 0.12 since it has not been released.

bytecode/instr.py
```diff
@@ -34,6 +34,37 @@ def const_key(obj):
...
+        "IS_OP",
+        "CONTAINS_OP",
+        "FORMAT_VALUE",
+        "MAKE_FUNCTION",
+        "IMPORT_NAME",
+        # technically, these three do not push back, but leave the container
+        # object on TOS
+        "SET_ADD",
+        "LIST_APPEND",
+        "MAP_ADD",
```
MatthieuDartiailh, line 64: Going through the list of opcodes again LOAD_ATTR should be listed

bytecode/instr.py
```diff
@@ -282,6 +313,27 @@ def stack_effect(self, jump=None):
         else:
             return dis.stack_effect(self._opcode, arg, jump=jump)
 
+    def pre_and_post_stack_effect(self, jump=None):
+
+        _effect = self.stack_effect(jump=jump)
```
MatthieuDartiailh, line 318: FOR_ITER needs also a special case since it adds to the stack based on TOS for a jump not taken and pop TOS when jumping

bytecode/cfg.py
```diff
@@ -140,7 +141,7 @@ def update_size(delta, size, maxsize):
         # jump is taken.
         if instr.has_jump():
             taken_size, maxsize = update_size(
-                instr.stack_effect(jump=True), size, maxsize
+                *instr.pre_and_post_stack_effect(jump=True), size, maxsize
```
MatthieuDartiailh, line 144: It could be nice while we stabilize this to have a way to bypass this in case we get the computation wrong (using an extra keyword to` _compute_stack_size`)

MatthieuDartiailh (changes requested):
Going through the opcodes again I found a couple of cases I missed. Sorry for that. Also since I feel we may still get it wrong it may be interesting to provide a way to skip this check through an extra keyword argument.

Thanks for your work on this !

bytecode/cfg.py
Z-Shang: I've made it a kw only argument :)

bytecode/instr.py
Z-Shang: actually I don't think this needs special consideration since the stack effect of  `FOR_ITER` is push 1 when no jump and pop 1 when jump, since the stack effect of non-pushing backs are all returned as "pre effect", this should be just fine.

MatthieuDartiailh: But in the first case you need something of the stack so for your analysis to hold you would need a pre effect of -1 and a post effect of 2, no ?

Z-Shang: ah yea, that makes sense then, thank you for the clarification

bytecode/cfg.py
```diff
@@ -141,7 +141,11 @@ def update_size(pre_delta, post_delta, size, maxsize):
         # jump is taken.
         if instr.has_jump():
             taken_size, maxsize = update_size(
```
MatthieuDartiailh, line 143: I would do the following to not have to pass check_pre_and_post to the instruction.
```python
effects = instr.pre_and_post_stack_effect(jump=True) if check_pre_and_post else (0, instr.stack_effect(jump=True)
taken_size, maxsize = update_size(*effects, size, maxsize)
```
and something similar for the jump False case.

bytecode/instr.py
```diff
@@ -313,9 +314,12 @@ def stack_effect(self, jump=None):
         else:
             return dis.stack_effect(self._opcode, arg, jump=jump)
 
-    def pre_and_post_stack_effect(self, jump=None):
-
+    def pre_and_post_stack_effect(self, jump=None, check_pre_and_post=True):
```
MatthieuDartiailh, line 317: I dislike the idea of this function not doing what its name suggests see my above suggestion.

bytecode/tests/test_bytecode.py
```diff
@@ -253,6 +253,62 @@ def test_to_code(self):
...
```
MatthieuDartiailh, line 284: This test could be parametrize for all other similar opcodes.

bytecode/instr.py
```diff
@@ -282,6 +314,33 @@ def stack_effect(self, jump=None):
...
+            # if the op pushes value back to the stack, then the stack effect given by dis.stack_effect
+            # actually equals pre + post effect, therefore we need -1 from the stack effect as a pre
+            # condition
+            return _effect - 1, 1
+        if _opname.startswith("UNPACK_"):
+            # Instr(UNPACK_* , n) pops 1 and pushes n
+            # _effect = n - 1
+            # hence we return -1, _effect + 1
+            return -1, _effect + 1
+        if _opname == "FOR_ITER" and not jump:
```
MatthieuDartiailh, line 337: Having a test for this case would be nice

bytecode/instr.py
```diff
@@ -282,6 +314,33 @@ def stack_effect(self, jump=None):
...
+        # To compute pre size and post size to avoid segfault cause by not enough stack element
+        _opname = _opcode.opname[self._opcode]
+        if _opname.startswith("DUP_TOP"):
+            return _effect * -1, _effect * 2
+        if _pushes_back(_opname):
+            # if the op pushes value back to the stack, then the stack effect given by dis.stack_effect
+            # actually equals pre + post effect, therefore we need -1 from the stack effect as a pre
+            # condition
+            return _effect - 1, 1
+        if _opname.startswith("UNPACK_"):
```
MatthieuDartiailh, line 332: Having a test for this case would be nice

bytecode/tests/test_bytecode.py
```diff
@@ -253,6 +253,62 @@ def test_to_code(self):
...
+            code.compute_stacksize()
+
+    def test_not_enough_dup(self):
+        code = Bytecode()
+        code.first_lineno = 1
+        code.extend([Instr("LOAD_CONST", 1), Instr("DUP_TOP_TWO")])
+        with self.assertRaises(RuntimeError):
+            code.compute_stacksize()
+
+    def test_not_enough_rot(self):
```
MatthieuDartiailh, line 305: This test could be parametrized too to test all rot.

MatthieuDartiailh (changes requested):
Couple minor things, also tests and documentation could be updated. Given the number of iterations I am putting you through let me know if you want me to help with anything.

bytecode/cfg.py
Z-Shang: yea, that was also my first thought, but it seems that the coverage test would fail, that's why I changed it to this way

bytecode/instr.py
Z-Shang: I think I'd like some help on constructing a test case for this one, it would be great if you can help :)

MatthieuDartiailh:
Sorry I have been awfully busy these past weeks. I will try to add the test you asked me to.

Z-Shang:
No worries, I'm also busy as hell these days :(

MatthieuDartiailh:
This looks really good. I will add a bit to the documentation in the coming week and then merge. Sorry again for the slow process I just moved from the USA back to France and I have been quite busy as a consequence.

Z-Shang:
Pas de souci et prends ton temps :)

MatthieuDartiailh:
I rebased to get tests running on 3.9. Once CIs run I will merge shortly. Thanks again for your contribution !

MatthieuDartiailh (approved):

PR #75, opened 2021-04-10
feat: add Python 2 support

P403n1x87:
I understand there is no interest in Python 2 support these days, so feel free to close this PR. I just thought of opening it for anybody who needs support for Python 2.


MatthieuDartiailh:
As mentionned in another thread the next release will drop support for Python 3.5 so I won't be adding support for Python 2 in particular now that we are past its EOL.

PR #76, opened 2021-04-11
chore: add Python support updates to the README

P403n1x87:
Future releases of this package drop support for versions of Python earlier than 3.6. Also inform that the latest version to support Python 3.5 is 0.12.0 and add a link to the Python 2 backport.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,8 +33,8 @@ bytecode
 
 Install bytecode: ``python3 -m pip install bytecode``. It requires Python 3.6
 or newer. The latest release that supports Python 3.5 is 0.12.0. For Python 2.7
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Thanks. I do not know how long you plan to maintain your Python 2 fork so could you rephrase your sentence to make it clear that this project does not make any guarantee about yours. Something like the following would be sufficient: " For Python 2 support, you can have a look at "

P403n1x87:
@MatthieuDartiailh Thanks. Given that Python 2 is dead now, I do not foresee a big maintenance burden from a Python 2-only library. It might just be the occasional bug fix, so hopefully, it'll be manageable and there won't be the need to prepare too many releases.

PR #79, opened 2021-05-06
cis: start testing on python 3.10.b1

MatthieuDartiailh:


bytecode/cfg.py
```diff
@@ -219,8 +222,20 @@ def compute_stacksize(self, *, check_pre_and_post=True):
             block.seen = False
             block.startsize = -32768  # INT_MIN
 
+        # Starting with Python 3.10, generator and coroutines start with one object
+        # on the stack (None, anything is an error).
+        initial_stack_size = 0
+        if sys.version_info >= (3, 10) and self.flags & (
```
P403n1x87, line 228: Is the flag check correct? AND has higher precedence than OR so I would've thought that you needed parentheses around those flags.

P403n1x87: Ah! I couldn't see the `(` from my tablet 🙁 . Nevermind!

MatthieuDartiailh: No problem.

PR #85, opened 2021-09-10
Pack linetable properly for Python 3.10. Fixes #84

### LINKED ISSUE #84: Bytecode doesn't properly set linenumbers (in corner case in Python 3.10rc1)
For some context, I use bytecode to set programmatic breakpoints in the pydev debugger and while using it with Python 3.10, I found a corner case where the line isn't being properly mapped back to bytecode.

-- note that it works in Python 3.9 and it also works if the target file is a bit different -- say with less lines or with smaller lines -- so, it seems I got extremely lucky that I have a test case which got into this situation...

I'm attaching a test case which shows the issue. What the test does is load the code for a function (in this case `long_lines_example.long_lines`) and change the 2nd line of that function to include bytecode which does something as the code below and generates back a code object:

```
    label = Label()
    return [
        # -- if _pydev_needs_stop_at_break():
        Instr("LOAD_CONST", _pydev_needs_stop_at_break, lineno=stop_at_line),
        Instr("LOAD_CONST", stop_at_line, lineno=stop_at_line),
        Instr("CALL_FUNCTION", 1, lineno=stop_at_line),
        Instr("POP_JUMP_IF_FALSE", label, lineno=stop_at_line),

        #     -- _pydev_stop_at_break()
        #
        # Note that this has line numbers -1 so that when the NOP just below
        # is executed we have a spurious line event.
        Instr("LOAD_CONST", _pydev_stop_at_break, lineno=stop_at_line - 1),
        Instr("LOAD_CONST", stop_at_line, lineno=stop_at_line - 1),
        Instr("CALL_FUNCTION", 1, lineno=stop_at_line - 1),
        Instr("POP_TOP", lineno=stop_at_line - 1),

        # Put NOP in new line so that Python given a line event for the debugger.
        Instr("NOP", lineno=stop_at_line),
        label,
    ]
```

This works well enough in general, but for this specific use case in Python 3.10 it's not putting the `NOP` in a new line.

[test_bytecode_line_not_correct.zip](https://github.com/MatthieuDartiailh/bytecode/files/7139096/test_bytecode_line_not_correct.zip)

Any ideas on what may be wrong there?

fabioz:


bytecode/concrete.py
```diff
@@ -280,11 +280,7 @@ def _assemble_lnotab(first_lineno, linenos):
         return b"".join(lnotab)
 
     @staticmethod
-    def _pack_linetable(doff, dlineno):
-        linetable = []
-        while doff > 254:
-            linetable.append(b"\xfe\x00")
-            doff -= 254
+    def _pack_linetable(doff, dlineno, linetable):
```
MatthieuDartiailh, line 283: I would prefer for linetable to be the first argument TBH.
```suggestion
    def _pack_linetable(linetable, doff, dlineno):
```

bytecode/concrete.py
```diff
@@ -280,11 +280,7 @@ def _assemble_lnotab(first_lineno, linenos):
...
 
         while dlineno < -127:
```
MatthieuDartiailh, line 285: Line may be too long with comment...

```suggestion
        # Ensure linenos are between -126 and +126, by using 127 lines jumps with a 0 byte offset
        while dlineno < -127:
```

bytecode/concrete.py
```diff
@@ -294,21 +290,33 @@ def _pack_linetable(doff, dlineno):
             linetable.append(struct.pack("Bb", 0, 127))
             dlineno -= 127
 
+        if doff > 254:
```
MatthieuDartiailh, line 293: Check line length

```suggestion
        # Ensure offsets are less than 255.
        # If an offset is larger, we first mark the line change with an offset of 254
        # then use as many 254 offset with no line change to reduce the offset to 
        # less than 254.
        if doff > 254:
```

bytecode/concrete.py
```diff
@@ -317,12 +325,12 @@ def _assemble_linestable(self, first_lineno, linenos):
             doff = offset - old_offset
             old_offset = offset
 
-            linetable.extend(self._pack_linetable(doff, old_dlineno))
+            self._pack_linetable(doff, old_dlineno, linetable)
```
MatthieuDartiailh, line 328:
```suggestion
            self._pack_linetable(linetable, doff, old_dlineno)
```

bytecode/concrete.py
```diff
@@ -317,12 +325,12 @@ def _assemble_linestable(self, first_lineno, linenos):
...
             old_dlineno = dlineno
 
         # Pack the line of the last instruction.
         doff = offset + i_size - old_offset
-        linetable.extend(self._pack_linetable(doff, old_dlineno))
+        self._pack_linetable(doff, old_dlineno, linetable)
```
MatthieuDartiailh, line 333:
```suggestion
        self._pack_linetable(linetable, doff, old_dlineno)
```

bytecode/tests/long_lines_example.py
```diff
@@ -0,0 +1,85 @@
...
+    a12 = 1 if a > 1 else 2
+    a13 = 1 if a > 1 else 2
+
+    for i in range(1):
+        if a > 0:
+            stream.write("111\n")
+            # a = 1
+        else:
+            stream.write("222\n")
+    return b
```
MatthieuDartiailh, line 74: Are you using this ?

MatthieuDartiailh (changes requested):
Thanks for looking into this !

I suggested a couple of changes to add more comments and a question about a test file. Also can you add an entry to the changelog ?

MatthieuDartiailh:
ping @fabioz

MatthieuDartiailh:
And sorry for forgetting to enable CIs.

bytecode/concrete.py
fabioz: Ok

bytecode/tests/long_lines_example.py
fabioz: I changed the test file to remove what's not being used.

fabioz:
@MatthieuDartiailh I've just done the changes and updated the branch. Please let me know if you think something else needs to be changed.

fabioz:
It seems it was failing due to black. I just reformatted and submitted again (the failures related to running with coverage seem unrelated).

MatthieuDartiailh:
The codecov issue is hopefully transient, I am not sure why black is still complaining

fabioz:
I just did one more pass at black (I had an older version of it and just upgraded), so, hopefully it should pass now.

MatthieuDartiailh:
I fixed the last linting issue and I will merge shortly. I saw your email to Python-dev so I will wait to see what comes out of it before cutting the next release.

MatthieuDartiailh (approved):

PR #92, opened 2022-02-10
Make test imports relative to tests package

juliangilbey:
The test suite currently has lots of imports of the form `from bytecode.tests import ...`.  This means that the tests can only be run when they are present together with the bytecode package itself, not independently.  This patch replaces these imports with relative imports: `from . import ...`.  All of the tests continue to run successfully with this patch when run in the current way, but they also work when the `tests` directory is copied elsewhere and the `tests` directory is removed from the installed package (for the tests are not needed for the bytecode package to work).  The only test which does not work when the tests are run on the installed package is `test_version`, because that requires the presence of `setup.py`, but that is a minor issue.

### DIFF
--- a/bytecode/tests/test_bytecode.py
+++ b/bytecode/tests/test_bytecode.py
@@ -3,7 +3,7 @@ import sys
 import textwrap
 import unittest
 from bytecode import Label, Instr, FreeVar, Bytecode, SetLineno, ConcreteInstr
-from bytecode.tests import TestCase, get_code
+from . import TestCase, get_code
 
 
 class BytecodeTests(TestCase):
--- a/bytecode/tests/test_cfg.py
+++ b/bytecode/tests/test_cfg.py
@@ -13,7 +13,7 @@ from bytecode import (
     ControlFlowGraph,
 )
 from bytecode.concrete import OFFSET_AS_INSTRUCTION
-from bytecode.tests import disassemble as _disassemble, TestCase
+from . import disassemble as _disassemble, TestCase
 
 
 def disassemble(
--- a/bytecode/tests/test_code.py
+++ b/bytecode/tests/test_code.py
@@ -1,7 +1,7 @@
 import unittest
 
 from bytecode import ConcreteBytecode, Bytecode, ControlFlowGraph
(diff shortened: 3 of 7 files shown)


P403n1x87:
I'd rather move the `tests` folder one level up, alongside `bytecode`. Would that still work?

MatthieuDartiailh:
My opinion on the matter tends to vary on a per project basis, but for such a small library, I like shipping the tests. In addition, if tests were at the top level we could not test against an installed version using pytest due to the presence of an `__init__.py` which we cannot remove. Otherwise, we would need a src directory but that would mess up the history.

So I am fine with the changes proposed but I am against moving the folder.

juliangilbey:
You can certainly test against an installed version using pytest, even with `__init__.py` present.  That is exactly what I am doing in my provisional Debian package of bytecode: I install bytecode without the tests directory, then copy the tests directory (with this patch applied) into a temporary directory and run `python -m pytest -k "not test_version" tests` in that temporary directory.  It works perfectly.

MatthieuDartiailh:
I was considering the simple workflow of just being able to invoke pytest from the root without any complicated dance.

Could you elaborate as to why you find the included tests bothering ?

Also, sorry for the rather late response.

juliangilbey:
Hi @MatthieuDartiailh!
There are two separate things here:
* Including the tests in the package.  They do no harm, though most end users won't care about them, so it's probably cleaner not to include them.
* Being able to run the tests independently.  Part of the purpose of the tests (and thanks for writing them, by the way!) is to check that the package works correctly.  A script that uses the `bytecode` package will do something like `import bytecode`, so the tests would be somewhat better if they test this usage pattern.  A test that can only be run from within the package itself does not fully test the package from a user's perspective.  That is what I am suggesting in this PR: ensure that the tests can be run if installed in a random location not inside the package itself.

I hope that makes some sense!

MatthieuDartiailh:
I am fine moving to relative imports, and I will merge shortly. I understand your use. I agree most users won't need the tests, but they get access to utilities used to write them, which can be of interest to them. So at this time I won't move the tests themselves.

Thanks for your contribution.

PR #94, opened 2022-02-23
fix: exclude EXC_MATCH from Compare enum in Python>=3.9

### LINKED ISSUE #93: Remove `Compare.EXC_MATCH` on Python>=3.9
Python 3.9 introduced the `JUMP_IF_NOT_EXC_MATCH` opcode, and the use of `Compare.EXC_MATCH` produces bad code that results in
~~~
TypeError: 'UH��H��]�l�' not supported between instances of 'type' and 'type'
~~~
therefore I would like to suggest that the attribute `Compare.EXC_MATCH` be defined only under the condition `sys.version_info[:2] < (3, 9)`

P403n1x87:
Python 3.9 introduced `JUMP_IF_NOT_EXC_MATCH` and removed the exception match from the set of compare operations. This change ensures that `EXC_MATCH` is not exported by the Compare enum from Python 3.9 onward. The new opcode should be used instead.

Fixes #93.

### DIFF
--- a/bytecode/instr.py
+++ b/bytecode/instr.py
@@ -19,7 +19,8 @@ class Compare(enum.IntEnum):
     NOT_IN = 7
     IS = 8
     IS_NOT = 9
-    EXC_MATCH = 10
+    if sys.version_info < (3, 9):
+        EXC_MATCH = 10
 
 
 UNSET = object()
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,15 @@
 ChangeLog
 =========
 
+23/02/2022: Version 0.13.1
+--------------------------
+
+Bugfixes:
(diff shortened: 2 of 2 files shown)


MatthieuDartiailh:
Could you add a changelog entry ?

P403n1x87:
@MatthieuDartiailh done

MatthieuDartiailh:
Thanks. I will try to release soon but it may not be today and I will update the release date in consequence.

MatthieuDartiailh:
Actually do you need a release quickly @P403n1x87 ?

I would like to overall the build system to use PEP 517 and also address #95 if possible before making a release but it may take me some time ?

P403n1x87:
Thanks, no rush from my side.

PR #100, opened 2022-03-29
drop support for Python 3.6 and 3.7

MatthieuDartiailh:


doc/changelog.rst
```diff
@@ -15,6 +15,7 @@ Bugfixes:
 Maintenance:
 
 - Make the install process PEP517 compliant PR #97
+- Drop support for Python 3.6 and 3.7 PR #100
```
P403n1x87, line 18: This probably requires a minor version bump, so 0.14.0 instead of 0.13.1. It would be great if the current support is also summarised in the README.

MatthieuDartiailh: Addressed in 7b042323

PR #102, opened 2022-05-24
fix: exclude IS, IS_NOT, IN, NOT_IN from Compare enum in Python>=3.9

### LINKED ISSUE #101: The `Compare` enum is broken starting from Python 3.9
~~~
Python 3.9.7 (default, Sep  3 2021, 12:37:55)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from dis import cmp_op
>>> len(cmp_op)
6
>>> cmp_op
('<', '<=', '==', '!=', '>', '>=')
~~~
The [`in`](https://docs.python.org/3.9/library/dis.html#opcode-CONTAINS_OP) and [`is`](https://docs.python.org/3.9/library/dis.html#opcode-CONTAINS_OP) operators have dedicated opcodes (with an argument for the inverse) and they should be removed from `Compare`.

P403n1x87:
Python 3.9 introduced the new opcodes `CONTAINS_OP` and `IS_OP` to perform `in` and `is` comparisons, instead of dedicated values for the argument to the `COMPARE_OP` instruction.

Fixes #101.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,10 @@ Bugfixes:
 - Removed ``EXC_MATCH`` from the ``Compare`` enumeration starting with Python
   3.9. The new ``JUMP_IF_NOT_EXC_MATCH`` opcode should be used instead.
 
+- Removed ``IN``, ``NOT_IN``, ``IS``, ``NOT_IS`` from the ``Compare``
+  enumeration starting with Python 3.9. The new ``CONTAINS_OP`` and ``IS_OP``
+  opcodes should be used instead.
+
 Maintenance:
 
 - Make the install process PEP517 compliant PR #97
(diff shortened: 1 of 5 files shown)


MatthieuDartiailh:
Thanks for the patch. Would be possible to add a test to avoid future regression ?

P403n1x87:
Yep, I'll add some tests as soon as I can. It would be great if the tests also run with every PR as I just discovered that some peephole optimisation tests are not passing with the new change and need to be adapted (as a follow up, it might be good to actually run the optimised code rather than just compare it to the expected abstract result).

MatthieuDartiailh:
CIs should run on PR yes. I must have made a configuration mistake for it not to be the case or I forgot to update it after moving to the src organization.

P403n1x87:
@MatthieuDartiailh this should be ready for review now

MatthieuDartiailh:
@P403n1x87 Could you rebaseon main ? I fixed the CI issues.

P403n1x87:
Oh I thought it was now possible to do that directly from the PR by those with merge permissions? Anyway I'll rebase this.

PR #108, opened 2022-10-11
Correctly ignore ellipses in overloads

SaladDais:
Follow-up for 6e6e38c2dc33bd16a1a4c787c6d5f82b6221b600

`exclude_lines` is a list of regexes, so `...` unescaped will actually exclude most lines from coverage analysis.
At the moment codecov is getting empty coverage data because of this exclusion rule.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -18,4 +18,4 @@ exclude_lines =
     pass
 
     # Don't complain about ellipsis in overload
-    ...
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh (approved):
I think I have the same fix in the 3.11 branch bur better to get it early. Thanks.

PR #117, opened 2023-03-27
Fix for extended_arg NOPs

EmmaJaneBonestell:
Re: https://github.com/MatthieuDartiailh/bytecode/issues/116


MatthieuDartiailh:
Your test need to also cover Python 3.11. Otherwise it looks good.

EmmaJaneBonestell:
Whoops. Since the args to CodeType aren't documented well, I missed that. Should all work now.

Issue #118, opened 2023-04-14
except Exception as e fails on Python 3.11


bburan:
```
import textwrap
from bytecode import Bytecode, ControlFlowGraph


source = '''
try:
    pass
except Exception as e:
    pass
'''
source = textwrap.dedent(source).strip()
code = compile(source, '<string>', 'exec')
bytecode = Bytecode.from_code(code)
cfg = ControlFlowGraph.from_bytecode(bytecode)
cfg.to_bytecode()
```
Gives the following traceback:
```
Traceback (most recent call last):
  File "c:\Users\lbhb\projects\psi-nafc\src\test_bytecode.py", line 15, in <module>
    cfg.to_bytecode()
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\bytecode\cfg.py", line 992, in to_bytecode
    byt_te.entry.stack_depth = min(
                               ^^^^
TypeError: '<' not supported between instances of '_UNSET' and '_UNSET'
```

MatthieuDartiailh:
Thanks for the report !

PR #119, opened 2023-04-18
Fix for #118

MatthieuDartiailh:
@bburan can you confirm it fixes your issue (and with enaml too)

### DIFF
--- a/src/bytecode/cfg.py
+++ b/src/bytecode/cfg.py
@@ -989,9 +989,14 @@ class ControlFlowGraph(_bytecode.BaseBytecode):
                                 entry.target is instr.target
                                 and entry.push_lasti == instr.push_lasti
                             ):
-                                byt_te.entry.stack_depth = min(
-                                    entry.stack_depth, instr.stack_depth
-                                )
+                                # If the exception table is in a dead code portion
+                                # simply use a size of zero
+                                if entry.stack_depth is UNSET:
+                                    byt_te.entry.stack_depth = 0
+                                else:
+                                    byt_te.entry.stack_depth = min(
+                                        entry.stack_depth, instr.stack_depth
+                                    )
                                 try_begins[instr] = byt_te.entry
                                 continue
                         assert isinstance(new, TryBegin)
--- a/tests/test_cfg.py
+++ b/tests/test_cfg.py
@@ -4,6 +4,7 @@ import contextlib
 import inspect
 import io
 import sys
+import textwrap
 import types
 import unittest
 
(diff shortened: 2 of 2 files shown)


bburan:
I get the following:

```
(psi-nafc) c:\Users\lbhb\projects\psi-nafc\src\psilbhb>enaml-run test.enaml
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Scripts\enaml-run.exe\__main__.py", line 7, in <module>
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\runner.py", line 44, in main
    code = EnamlCompiler.compile(ast, enaml_file_path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_compiler.py", line 179, in compile
    return compiler.visit(node)
           ^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_ast.py", line 418, in visit
    result = visitor(node, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_compiler.py", line 207, in visit_Module
    return cg.to_code()
           ^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\code_generator.py", line 118, in to_code
    return bc_code.to_code()
           ^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\bytecode\bytecode.py", line 305, in to_code
    cfg = _bytecode.ControlFlowGraph.from_bytecode(self)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\bytecode\cfg.py", line 832, in from_bytecode
    assert active_try_begin
AssertionError
```
The test file is:
```
try:
    pass
except Exception as e:
    pass
```
It can be run with `enaml-run test.enaml` (yes, it doesn't have a Main in it, but that's OK since it still triggers the error).

MatthieuDartiailh:
If you can it would be great yes. It seems your real issue is not related to dead code only and I would like to understand it better.

MatthieuDartiailh:
ping @bburan

bburan:
I was trying to find the original code that triggered this, but since I did not need the "as" part of the exception, which is what was triggering this error, I had already removed it to work around this issue. However, the general gist is this:
```
import logging
log = logging.getLogger(__name__)
try:
    import backend_1
    backend_1_available = True
except ImportError as e:
    log.exception(e)
    backend_1_available = False
```
It's useful to capture and log the error message (since there are different reasons one may get an ImportError). This was at the top of my enaml file and was failing. The dead code example I gave you was a minimal working example, but this particular example is something you might expect to encounter in the real world. You may want to modify the behavior or layout of an Enaml widget/application based on whether backend_1_available is True or False. Now, we could definitely put this in a Python file and import the variable from the file; however, the whole point of Enaml is that any Python code should be valid, so users would end up getting a confusing error message if they do a "try/except ... as e" in their Enaml file.

MatthieuDartiailh:
Thanks for the actual use case. This is not only dead code so I definitively need to fix bytecode to support it.

MatthieuDartiailh:
So from the actual use case I was able to confirm that the issue does not stem from bytecode. The enaml compiler is doing something invalid and it needs to be fixed. This PR could nonethless be valuable so I will merge it and keep working on the enaml side.

PR #122, opened 2023-05-28
Python 3.12 support

MatthieuDartiailh:


MatthieuDartiailh:
Just a quick update:
- most tests are now passing
- some tests should probably be added for new behavior in Python 3.12
- stack size is currently broken for all cases involving `async with`

MatthieuDartiailh:
Framework tests have detected an issue with the computation of jumps. It may be related to `FOR_ITER` but does not appear as simple as the fix for `SEND`....

MatthieuDartiailh:
A single test in the framework test and does not look bytecode related.

PR #129, opened 2023-09-25
chore: handle LOAD_METHOD opcode in CPython 3.12

### LINKED ISSUE #128: `LOAD_METHOD` is still a valid opname but doesn't seem to be fully supported
Porting some code to 3.12 and I get
```
ddtrace/internal/wrapping/__init__.py:520: in wrap
    f.__code__ = code.to_code()
.riot/venv_py3120/lib/python3.12/site-packages/bytecode/bytecode.py:312: in to_code
    bc = self.to_concrete_bytecode(
.riot/venv_py3120/lib/python3.12/site-packages/bytecode/bytecode.py:327: in to_concrete_bytecode
    return converter.to_concrete_bytecode(
.riot/venv_py3120/lib/python3.12/site-packages/bytecode/concrete.py:1401: in to_concrete_bytecode
    self.concrete_instructions()
.riot/venv_py3120/lib/python3.12/site-packages/bytecode/concrete.py:1275: in concrete_instructions
    self.required_caches = c_instr.use_cache_opcodes()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <LOAD_METHOD arg=1 location=InstrLocation(lineno=105, end_lineno=None, col_offset=None, end_col_offset=None)>

    def use_cache_opcodes(self) -> int:
        print(self._opcode)
        return (
            # Not supposed to be used but we need it
>           dis._inline_cache_entries[self._opcode]  # type: ignore
            if sys.version_info >= (3, 11)
            else 0
        )
E       IndexError: list index out of range

.riot/venv_py3120/lib/python3.12/site-packages/bytecode/concrete.py:183: IndexError
```
so it looks like the `LOAD_METHOD` opcode is not supported by 0.15.0. However, `LOAD_ATTR` should provide a workaround by setting the first argument of the tuple to `True`.

P403n1x87:
Allow handling the `LOAD_METHOD` pseudo-instruction in CPython 3.12 by converting it to the equivalent `LOAD_ATTR` instruction.

Resolves #128.

### DIFF
--- a/src/bytecode/concrete.py
+++ b/src/bytecode/concrete.py
@@ -1143,6 +1143,17 @@ class _ConvertBytecodeToConcrete:
         free_instrs: List[int] = []
 
         for instr in self.bytecode:
+            if sys.version_info >= (3, 12) and isinstance(instr, Instr):
+                if instr.name == "LOAD_METHOD":
+                    # The LOAD_METHOD opcode has become a pseudo-instruction in
+                    # Python 3.12
+                    instr = Instr(
+                        "LOAD_ATTR",
+                        (True, instr.arg),
+                        lineno=instr.lineno,
+                        location=instr.location,
+                    )
+
             # Enforce proper use of CACHE opcode on Python 3.11+ by checking we get the
             # number we expect or directly generate the needed ones.
             if isinstance(instr, Instr) and instr.name == "CACHE":
--- a/src/bytecode/instr.py
+++ b/src/bytecode/instr.py
@@ -805,6 +805,9 @@ class Instr(BaseInstr[InstrArg]):
 
         elif opcode in _opcode.haslocal or opcode in _opcode.hasname:
             if name in BITFLAG_INSTRUCTIONS:
+                if isinstance(arg, str):
+                    # Assume the bitflag is not set
+                    arg = (False, arg)
                 if not (
                     isinstance(arg, tuple)
                     and len(arg) == 2
--- a/tests/test_instr.py
+++ b/tests/test_instr.py
@@ -191,7 +191,6 @@ class InstrTests(TestCase):
 
         # Instructions using a bitflag in their oparg
         for name in BITFLAG_INSTRUCTIONS:
-            self.assertRaises(TypeError, Instr, name, "arg")
             self.assertRaises(TypeError, Instr, name, ("arg",))
             self.assertRaises(TypeError, Instr, name, ("", "arg"))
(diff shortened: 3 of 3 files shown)


src/bytecode/instr.py
```diff
@@ -805,6 +805,9 @@ def _check_arg(self, name: str, opcode: int, arg: InstrArg) -> None:
 
         elif opcode in _opcode.haslocal or opcode in _opcode.hasname:
             if name in BITFLAG_INSTRUCTIONS:
+                if isinstance(arg, str):
+                    # Assume the bitflag is not set
+                    arg = (False, arg)
```
P403n1x87, line 810: This is a breaking change from what currently documented

src/bytecode/concrete.py
```diff
@@ -1143,6 +1143,17 @@ def concrete_instructions(self) -> None:
         free_instrs: List[int] = []
 
         for instr in self.bytecode:
+            if sys.version_info >= (3, 12) and isinstance(instr, Instr):
+                if instr.name == "LOAD_METHOD":
```
P403n1x87, line 1147: Would there be a better place to handle these sorts of transformations? Just thinking of ways of keeping this tidy in case of future similar changes.

MatthieuDartiailh:
I am extremely hesitant to allow this kind of things. In particular I fear that "legalizing" the bytecode when converting to concrete may be too late since we need a CFG to compute the stack size and I would not bet on dis.stack_effect being accurate for pseudo instructions. 

This and your request about TryBegin/End to me sound more like we could have a more abstract builder for bytecode that we could then safely compile before doing any computation on it. Food for though.

P403n1x87:
I've had this idea for a while now and finally had a chance to experiment with it https://github.com/P403n1x87/spasmlang. Perhaps these sorts of transformations can be handled at this higher level.

P403n1x87:
I'm closing this PR since my use cases are now covered by https://github.com/P403n1x87/spasmlang/pull/2.

PR #131, opened 2023-10-12
Fix spelling of INTRINSIC

juliangilbey:
There are 12 places in the repo where "INTRINSIC" has been misspelled as "INSTRINSIC".  Most of these are harmless as they are only in documentation, but there are 4 cases in `src/bytecode/instr.py` which would probably cause errors.  This PR fixes all 12 cases.


MatthieuDartiailh:
Can you add a changelog entry (under a new bugfix release) ? It is important since it touches some public enums.

juliangilbey:
I think this changelog does the job you want!  I haven't created a tag for the version number - I think this is the only way that the version number is available to the setup code.

MatthieuDartiailh:
That's perfect. I will try to address a couple other issues before tagging the release.

MatthieuDartiailh (approved):

PR #143, opened 2024-03-19
ci: improve code round-tripping

P403n1x87:
We improve the logic for round-trip re-compilation of code objects by listening for module code objects before they are executed. This way we can thoroughly recurse over nested code object and attain total coverage for better confidence.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,13 @@
 ChangeLog
 =========
 
+2024-05-28: Version 0.15.3
+--------------------------
+
+Bugfixes:
+
+- Ensure the correct management of TryBegin blocks in CFGs. PR #1xx
+
 2024-05-28: Version 0.15.2
 --------------------------
 
--- a/src/bytecode/cfg.py
+++ b/src/bytecode/cfg.py
@@ -358,12 +358,13 @@ class _StackSizeComputer:
                 if instr.is_uncond_jump():
                     # Check for TryEnd after the final instruction which is possible
                     # TryEnd being only pseudo instructions
-                    if te := self.block.get_trailing_try_end(i):
-                        if self._current_try_begin is None:
-                            self._current_try_begin = te.entry
-                        else:
-                            # TryBegin cannot be nested
-                            assert te.entry is self._current_try_begin, (te.entry, self._current_try_begin)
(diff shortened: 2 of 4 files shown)


MatthieuDartiailh:
And sadly it breaks. 
It will likely take me some time before I can look into this though.

P403n1x87:
> And sadly it breaks. It will likely take me some time before I can look into this though.

No worries. I haven't come across any issues so these might be a few corner cases, and only on 3.11 https://github.com/MatthieuDartiailh/bytecode/actions/runs/8344140634/job/22835903388?pr=143

P403n1x87:
Maybe something to look into is why it seems to take about twice as long on 3.12 to recompile about the same number of code objects
```
Recompiled 11901 code objects in 0:00:48.426503
```

MatthieuDartiailh:
Compared to what ? 3.10 ?

To me that is not that surprising due to the introduction of exception tables in 3.11.

P403n1x87:
3.11 took 36s (I think we can assume it re-compiled all code objects without errors). 3.10 and earier are all in the ballpark of 25s. It makes sense for 3.11 and later versions to take more time because of the extra data structures, but the exception table is generally much smaller than the bytecode itself, so I don't know if that alone can really explain a 100% run time increase. I'll see if I can get some CPU profiling data

MatthieuDartiailh:
@P403n1x87 did you ever manage to generate the profiling data ?

P403n1x87:
@MatthieuDartiailh the tests on 3.13 pass without the re-compilation. It looks like the wrong arguments are being resolved. I have added

```python
        print("availabe_services", available_services)
```
and these are the results

## With recomp
```
unit/test_session.py availabe_services Session(region_name=<Mock name='mock.get_config_variable()' id='4875256224'>)
Favailabe_services Session(region_name=<Mock name='mock.get_config_variable()' id='4875263280'>)
```

## Without recomp
```
unit/test_session.py availabe_services ['good-resource']
.availabe_services ['good-resource']
```

P403n1x87:
@MatthieuDartiailh I have the feeling that the CFG needs a similar fix to #145
```
    assert te.entry is self._current_try_begin
```
WDYT?

MatthieuDartiailh:
Superseded by #149

Issue #144, opened 2024-05-16
Cannot decompile code with empty try block


P403n1x87:
Small reproducer: we add a `TryBegin` followed immediately by the closing `TryEnd`. The code re-compiles fine, but cannot be decompiled again

```python
def test_bytecode():
    import bytecode as b

    def foo():
        return 42

    bc = b.Bytecode.from_code(foo.__code__)
    label = b.Label()
    try_begin = b.TryBegin(label, push_lasti=True)
    bc[1:1] = [try_begin, b.TryEnd(try_begin), label]

    foo.__code__ = bc.to_code()

    assert foo() == 42

    bc = b.Bytecode.from_code(foo.__code__)
```
Result:
```console
________________________________ test_bytecode _________________________________

    def test_bytecode():
        import bytecode as b
    
        def foo():
            return 42
    
        bc = b.Bytecode.from_code(foo.__code__)
        label = b.Label()
        try_begin = b.TryBegin(label, push_lasti=True)
        bc[1:1] = [try_begin, b.TryEnd(try_begin), label]
    
        foo.__code__ = bc.to_code()
    
        assert foo() == 42
    
>       bc = b.Bytecode.from_code(foo.__code__)

tests/internal/test_wrapping.py:779: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.riot/venv_py3120/lib/python3.12/site-packages/bytecode/bytecode.py:283: in from_code
    return concrete.to_bytecode(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ConcreteBytecode instr#=2>, prune_caches = True
conserve_exception_block_stackdepth = False

    def to_bytecode(
        self,
        prune_caches: bool = True,
        conserve_exception_block_stackdepth: bool = False,
    ) -> _bytecode.Bytecode:
        # On 3.11 we generate pseudo-instruction from the exception table
    
        # Copy instruction and remove extended args if any (in-place)
        c_instructions = self[:]
        self._remove_extended_args(c_instructions)
    
        # Find jump targets
        jump_targets: Set[int] = set()
        offset = 0
        for c_instr in c_instructions:
            if isinstance(c_instr, SetLineno):
                continue
[...]

PR #145, opened 2024-05-17
fix: handle small try blocks

P403n1x87:
We make sure that we can handle small (and potentially empty) try blocks. These are blocks where the end offset is smaller or at most equal to the start offset. These blocks can occur when a single instruction that requires EXTENDED_ARG opcodes is wrapped in a try block. Original code objects seem to wrap the whole instruction, including the EXTENDED_ARG opcodes. The fix in this PR wraps just the instruction, leaving the extra EXTENDED_ARG opcodes outside the try block. This might not be a problem in general, but it is perhaps not ideal.

Addresses #144.

### DIFF
--- a/tests/test_bytecode.py
+++ b/tests/test_bytecode.py
@@ -712,6 +712,64 @@ class BytecodeTests(TestCase):
                 while callable(f := f()):
                     pass
 
+    def test_empty_try_block(self):
+        if sys.version_info < (3, 11):
+            self.skipTest("Exception tables were introduced in 3.11")
+
+        import bytecode as b
+
+        def foo():
+            return 42
+
+        code = Bytecode.from_code(foo.__code__)
+
+        try_begin = b.TryBegin(Label(), push_lasti=True)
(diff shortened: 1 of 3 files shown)


tests/test_bytecode.py
```diff
@@ -712,6 +712,64 @@ def recompile_code_and_inner(code):
...
+        code = Bytecode.from_code(foo.__code__)
+        foo.__code__ = code.to_code()
+
+        # Test that the function is still good
+        self.assertEqual(foo(), 42)
+
+        # Do another round trip
+        Bytecode.from_code(foo.__code__).to_code()
+
+    def test_try_block_around_extended_arg(self):
```
MatthieuDartiailh, line 744: Could you add a comment motivating this test for future reference ?

MatthieuDartiailh (changes requested):
Could you add a changelog entry and address my small comment and I will merge.

MatthieuDartiailh (approved):

PR #147, opened 2024-10-17
fix(cfg): ensure correct TryBegin management

P403n1x87:
We make sure that the TryBegin blocks are managed correctly in CFGs

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,13 @@
 ChangeLog
 =========
 
+2024-05-28: Version 0.15.3
+--------------------------
+
+Bugfixes:
+
+- Ensure the correct management of TryBegin blocks in CFGs. PR #1xx
+
 2024-05-28: Version 0.15.2
 --------------------------
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -6,7 +6,7 @@ ChangeLog
 
 Bugfixes:
 
-- Ensure the correct management of TryBegin blocks in CFGs. PR #1xx
+- Ensure the correct management of TryBegin blocks in CFGs. PR #147
 
 2024-05-28: Version 0.15.2
 --------------------------
--- a/src/bytecode/cfg.py
+++ b/src/bytecode/cfg.py
@@ -358,8 +358,11 @@ class _StackSizeComputer:
                     # Check for TryEnd after the final instruction which is possible
                     # TryEnd being only pseudo instructions
                     if te := self.block.get_trailing_try_end(i):
-                        # TryBegin cannot be nested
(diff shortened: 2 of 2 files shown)


P403n1x87:
I'm not quite sure this is the right fix. If there is no current TryBegin block set but we have a TryEnd, then perhaps we should have a TryBegin too?

P403n1x87:
Framework tests for CPython 3.11 are passing on #143 with this fix https://github.com/MatthieuDartiailh/bytecode/actions/runs/11385325044/job/31674923987?pr=143

doc/changelog.rst
```diff
@@ -1,6 +1,13 @@
 ChangeLog
 =========
 
+2024-05-28: Version 0.15.3
+--------------------------
+
+Bugfixes:
+
+- Ensure the correct management of TryBegin blocks in CFGs. PR #1xx
```
P403n1x87, line 9:
```suggestion
- Ensure the correct management of TryBegin blocks in CFGs. PR #147
```

MatthieuDartiailh:
This is a bit odd. Can you make a small test for this case. To me such a situation should not occur or may occur if the TryBegin itself is after the jump. A different fix could be to ignore a TryEnd if there is no known TryBegin.

P403n1x87:
> This is a bit odd. Can you make a small test for this case. To me such a situation should not occur or may occur if the TryBegin itself is after the jump. A different fix could be to ignore a TryEnd if there is no known TryBegin.

My gut feeling is that this is the same as #145 whereby the lookup for the entry would fail because the `TryBegin` was not created yet for small try blocks. I'll see if I can craft a reproducer for this.

P403n1x87:
I haven't had time to craft a reproducer or do more experiments.

> A different fix could be to ignore a TryEnd if there is no known TryBegin.

Perhaps one can skip `get_trailing_try_end` if there is no current TryBegin if that's what you're implying. Or perhaps one should ensure that `get_trailing_try_end` doesn't find a `TryBegin` before a `TryEnd`?

MatthieuDartiailh:
My idea was to skip. Maybe it's worth trying to see what the framework test finds.

P403n1x87:
Skipping seems to work too https://github.com/MatthieuDartiailh/bytecode/actions/runs/11416651416/job/31768093575?pr=143

MatthieuDartiailh:
Thanks for testing. Could you check that when we skip there is indeed a TryBegin after the end of the block ? 

I feel like all of this means something is broken in the generation of those TryBegin/TryEnd and that we should find the source rather than just addressing the symptoms.

MatthieuDartiailh:
Superseded by #149

PR #148, opened 2024-10-19
ci: fix linting

P403n1x87:
We fix the linting command by using the new ruff command. We also pin the tox linting environment dependencies to prevent the CI job from breaking again in the future.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,11 +20,10 @@ commands =
 [testenv:lint]
 basepython = python3
 deps=
-    ruff
-    mypy
+    ruff~=0.7.0
+    mypy~=1.12.0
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
I know that I won't remember to bump the linter versions so I am against pinning them (I find it easier to deal with the breakage than to remember to update the pin)

P403n1x87:
What is the need to update?

MatthieuDartiailh:
This now fixed on main so I will close. 

Personally I prefer to upgrade. People are putting a lot of effort in making the tools better. I consider that using the latest version is a way to recognize those efforts.

PR #149, opened 2024-10-27
Ci/improve code collection

MatthieuDartiailh:


MatthieuDartiailh:
@P403n1x87 I pulled your work #143 and after investigating a bit more the 3.11 failure came up with a marginally different solution (mostly a clearer explanation of why things happen). If you are fine with it I will merge and then proceed to 0.16.0 release with Python 3.13 support.

P403n1x87 (approved):

PR #157, opened 2025-01-14
Fix flag inference

### LINKED ISSUE #155: infer_flags mistakes coroutine function with comprehension as async generator
An async function with a comprehension and await causes is incorrectly marked as an async generator. 

```python
In [11]: async def bar():
    ...:     return 1
    ...: async def foo():
    ...:     x = [i for i in range(3)]
    ...:     await bar()
    ...:     return x
    ...: 

In [12]: code = bytecode.Bytecode.from_code(foo.__code__)

In [13]: code.flags
Out[13]: <CompilerFlags.OPTIMIZED|NEWLOCALS|COROUTINE: 131>

In [14]: code.update_flags()

In [15]: code.flags
Out[15]: <CompilerFlags.OPTIMIZED|NEWLOCALS|NOFREE|ASYNC_GENERATOR: 579>

```

MatthieuDartiailh:
Starting with Python 3.11, YIELD_VALUE appears in async function using await even if the function is not a generator. To identify if a YIELD_VALUE is associated with a generator like behavior one needs to inspect the argument of the following RESUME which was not done.

The new logic should be more robust and several round-trip test have been added to ensure we do not promote coroutine to async generator for no reason.

Closes #155

### DIFF
--- a/src/bytecode/flags.py
+++ b/src/bytecode/flags.py
@@ -107,7 +107,6 @@ def infer_flags(
     known_generator = False
     possible_generator = False
     instr_iter = iter(instructions)
-    print()
     for instr in instr_iter:
         if isinstance(
             instr,
--- a/src/bytecode/flags.py
+++ b/src/bytecode/flags.py
@@ -57,15 +57,10 @@ ASYNC_OPCODES = (
 YIELD_VALUE_OPCODE = _opcode.opmap["YIELD_VALUE"]
 GENERATOR_LIKE_OPCODES = (
     *((_opcode.opmap["YIELD_FROM"],) if not PY311 else ()),  # Removed in 3.11+
-    _opcode.opmap["RETURN_GENERATOR"],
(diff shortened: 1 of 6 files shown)


MatthieuDartiailh:
@frmdstryr can you have a look and confirm it does fix your issue ?

frmdstryr:
Yep it works here, thank you!

src/bytecode/flags.py
```diff
@@ -80,26 +98,74 @@ def infer_flags(
...
+
+    # Iterate over the instructions and inspect the arguments
+    is_concrete = isinstance(bytecode, _bytecode.ConcreteBytecode)
+    optimized = True
+    has_free = False if not is_concrete else bytecode.cellvars and bytecode.freevars
+    known_async = False
+    known_generator = False
+    possible_generator = False
+    instr_iter = iter(instructions)
+    print()
```
frmdstryr, line 110: Print needs removed

MatthieuDartiailh: Thanks

PR #163, opened 2025-04-17
cfg: fix stack depth of TryBegin when splitting across blocks.

arusuki:
When constructing the `ControlFlowGraph` from `Bytecode`, current implementation would split instructions into `BasicBlock` when encountering a **final** instruction. If the split point is inside a `TryBegin`/`TryEnd` pair, it will yield multiple copies of `TryBegin` and `TryEnd` instances for the generated blocks. However, currently the implementation does not set the `stack_depth` filed of the `TryBegin` copies, as in the [source](https://github.com/MatthieuDartiailh/bytecode/blob/0.16.2/src/bytecode/cfg.py#L875).

### DIFF
--- a/src/bytecode/cfg.py
+++ b/src/bytecode/cfg.py
@@ -871,9 +871,7 @@ class ControlFlowGraph(_bytecode.BaseBytecode):
                     #   the new one since the blocks are disconnected.
                     if last_instr.is_final() and temp:
                         old_block.append(TryEnd(try_begins[active_try_begin][-1]))
-                        new_tb = TryBegin(
-                            active_try_begin.target, active_try_begin.push_lasti
-                        )
+                        new_tb = active_try_begin.copy()
                         block.append(new_tb)
                         # Add this new TryBegin to the map to properly update
                         # the target.


arusuki:
As an example, function:

```python
def test():
    try:
        pass
    except RuntimeError:
        pass
    finally:
        for _ in range(10):
            try:
                print("1")
            except StopIteration:
                print("2")
```

with bytecode:
```text
...
296 RERAISE                  0
298 COPY                     3
300 POP_EXCEPT
302 RERAISE                  1

304 RERAISE                  0
306 COPY                     3
308 POP_EXCEPT
310 RERAISE                  1

ExceptionTable:
  ...
  296 to 296 -> 298 [4] lasti
  298 to 304 -> 306 [1] lasti

```

Current `dump_bytecode()` shows the result as follows:(**block20**, **block21** is created by splitting the code in offset 298 to 304, and **TryBegin15** is constructed from **TryBegin14**)

```text
...
block20:
    TryBegin 14 -> <block22> [1] last_i
      COPY 3
      POP_EXCEPT
      RERAISE 1
    TryEnd (14)

block21:
    TryEnd (8)
    TryBegin 15 -> <block22> [0] last_i
      RERAISE 0
    TryEnd (15)

block22:
    COPY 3
    POP_EXCEPT
    RERAISE 1

```

After change:


```text
...
block20:
    TryBegin 14 -> <block22> [1] last_i
      COPY 3
      POP_EXCEPT
      RERAISE 1
    TryEnd (14)

block21:
    TryEnd (8)
    TryBegin 15 -> <block22> [1] last_i
      RERAISE 0
    TryEnd (15)

block22:
    COPY 3
    POP_EXCEPT
    RERAISE 1

```

MatthieuDartiailh:
I understand your point but the example seems quite contrived. Reading the bytecode the TryBegin 15 looks like dead code to me (the previous reraise is final so one cannot get to 304).
But I may be missing something since the stackdepth of 1 for 298-304 looks wrong. I would expect 4 since we have a COPY 3 in the handler. What do you get with dis ?

arusuki:
The python version I'm using is CPython 3.11.11, the bytecode and exception table I'm giving here are all  from `dis.dis()`. 

First sorry for the somewhat complex example(I failed to make a simpler one that just incurs the split of TryBegin, so I just made a simplfied version of the function I'm working with), and here are my explanations.

* Actually, 304 is not dead code, in previous instructions there is a jump instruction that jumps right into the middle of a `TryBegin`/`TryEnd` pair(corresponding to the `298 to 304 -> 306 [1] lasti`entry in the exception table), **which I believe triggers the split of** `TryBegin`.

    ```
    ...
    block15:
          FOR_ITER <block21>   # block21 starts at offset 304
          -> block16
    ...
    ```
* The stack_depth of 1 is really interesting, according to the [CPython source](https://github.com/python/cpython/blob/3.11/Python/ceval.c#L5762), after expcetion is raised, the interpreter first reduces the stack size to the `stack_depth` according to the exception table entry, then push the `last_i`(if specified) and the exception value on the stack, the jumps to the handler for further evaluation. Here is the simplified version of the code(if you are interested).
   ```c
    int level, handler, lasti; // note the `level` is the stack_depth
    get_exception_handler(frame->f_code, offset, &level, &handler, &lasti)
    PyObject **new_top = _PyFrame_Stackbase(frame) + level;
    while (stack_pointer > new_top) {
          PyObject *v = POP();    // <--- first pop elements from stack so that stacksize == stack_depth
          Py_XDECREF(v);
     }
     PyObject *exc, *val, *tb;
     if (lasti) {
          // ... retrieve `lasti` from current frame
          PUSH(lasti);            // <-- push last_i
     }
     _PyErr_Fetch(tstate, &exc, &val, &tb);
     // ... excpetion normalization
     PUSH(val);                   // <-- push exception value

     ...
    ```
    As a result, when the `COPY 3` instruction is excuted, the real stack depth is 3 (including `last_i` and exception value), and element stack[-3] will be fetched. I believe you can find a lot of combination of `stack_depth=1` and `COPY 3` instructions in CPython 3.11 exception handling code.

MatthieuDartiailh:
Thanks for the very clear explanation and for refreshing my memory. I did not touch the exception table related code much since 3.11.

I will merge your fix. I am not exactly convinced it can have observable effects due to how we compute stack depth but it is more correct.

MatthieuDartiailh (approved):

PR #166, opened 2025-06-23
Python 3.14 support

MatthieuDartiailh:
Some backward incompatible changes are also introduce regarding:
- BINARY_OP that now requires a BinaryOp object as argument (for Instr)
- FORMAT_VALUE that now requires a FormatValue object as argument
This align the behavior of those instructions with others using an enum and hence requiring one.


MatthieuDartiailh:
3.14 tests are passing locally with 3.14.0b3

makslevental:
Hey just curious is this ready to land?

MatthieuDartiailh:
Actually tests are also missing for validation of the argument of the following instructions: BINARY_OP, FORMAT_VALUE, SPECIAL_METHOD, COMMON_CONSTANT, LOAD_SMALL_INT
For this you can extend: test_instr.py::InstrTest.test_invalid_arg

PR #167, opened 2025-07-25
add invalid arg tests for new bytecodes

makslevental:
tests for validation of the argument of the following instructions: BINARY_OP, FORMAT_VALUE, SPECIAL_METHOD, COMMON_CONSTANT, LOAD_SMALL_INT

### DIFF
--- a/tests/test_instr.py
+++ b/tests/test_instr.py
@@ -6,6 +6,7 @@ import unittest
 from bytecode import (
     UNSET,
     BasicBlock,
+    BinaryOp,
     CellVar,
     Compare,
     FreeVar,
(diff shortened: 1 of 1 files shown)


makslevental:
Let me know if this is what you had in mind

tests/test_instr.py
```diff
@@ -156,6 +159,41 @@ def test_invalid_arg(self):
         label = Label()
         block = BasicBlock()
 
+        if PY314:
```
MatthieuDartiailh, line 162: Since I define dedicated groups of opcode which are version dependent (such as BITFLAG_OPCODES) I would prefer for the tests to use those to avoid duplicating the version checks. Otherwise LGTM (and sorry for the long silence)

makslevental: Sure np I'll make the change

makslevental: okay I changed to iteration over the opcode groups (like I saw that you do for `BITFLAG_OPCODES`. It's not super "elegant" because I don't know if you plan to add more opcode to the respective groups...

MatthieuDartiailh:
It runs only for PR targeting main. I will merge as is and try to add the test regarding the disassembly of relying on FORMAT_VALUE and BUILD_INTERPOLATION.

PR #170, opened 2025-08-26
Fix the `next_block` attribute of the new block in `ControlFlowGraph.split_block`

BergLucas:
Hi,

First of all, thank you for maintaining this library, it is really useful and complete!

I'm doing this little PR because while I was refactoring Pynguin's instrumentation component (https://github.com/se2p/pynguin/pull/110), I encountered a small bug in the function `ControlFlowGraph.split_block`. When splitting a `BasicBlock`, I think that the `next_block` attribute of the newly created block should be assigned to the previous `next_block` attribute, but it is not.

If I'm not mistaken, the execution of the function should produce this result:

**Input:**
`BasicBlock1`.`next_block` = `BasicBlock2`

**Expected result:**
`BasicBlock1`.`next_block` = `BasicBlock3`
`BasicBlock3`.`next_block` = `BasicBlock2`

However, it produces this result instead:

**Actual result:**
`BasicBlock1`.`next_block` = `BasicBlock3`
`BasicBlock3`.`next_block` = `None`

Is this a bug, or is it expected behaviour?

Thank you in advance, and have a nice day!

### DIFF
--- a/tests/test_cfg.py
+++ b/tests/test_cfg.py
@@ -534,6 +534,8 @@ class BytecodeBlocksFunctionalTests(TestCase):
 
         label = code.split_block(code[0], 2)
         self.assertIs(label, code[1])
+        self.assertIs(code[0].next_block, label)
+        self.assertIs(label.next_block, None)
         self.assertBlocksEqual(
             code,
             [
@@ -546,6 +548,9 @@ class BytecodeBlocksFunctionalTests(TestCase):
 
         label2 = code.split_block(code[0], 1)
         self.assertIs(label2, code[1])
+        self.assertIs(code[0].next_block, label2)
+        self.assertIs(label2.next_block, label)
+        self.assertIs(label.next_block, None)
         self.assertBlocksEqual(
             code,
             [Instr("LOAD_SMALL_INT" if PY314 else "LOAD_CONST", 1, lineno=1)],
@@ -567,6 +572,8 @@ class BytecodeBlocksFunctionalTests(TestCase):
         # split at the end of the last block requires to add a new empty block
         label = code.split_block(code[0], 2)
         self.assertIs(label, code[1])
+        self.assertIs(code[0].next_block, label)
+        self.assertIs(label.next_block, None)
         self.assertBlocksEqual(
             code,
             [
@@ -581,6 +588,8 @@ class BytecodeBlocksFunctionalTests(TestCase):
         # add a new block
         label = code.split_block(code[0], 2)
         self.assertIs(label, code[1])
+        self.assertIs(code[0].next_block, label)
+        self.assertIs(label.next_block, None)
         self.assertBlocksEqual(
             code,
             [
@@ -596,6 +605,7 @@ class BytecodeBlocksFunctionalTests(TestCase):
         # FIXME: is it really useful to support that?
         block = code.split_block(code[0], 0)
         self.assertIs(block, code[0])
+        self.assertIs(code[0].next_block, None)
         self.assertBlocksEqual(
             code,
             [


MatthieuDartiailh:
I will need a bit more time to double check things but this makes sense to me. Could you modify a test to enforce this ?

BergLucas:
Here you go, I've modified the few tests about the `split_block` method to check that it works in each case.

MatthieuDartiailh:
Indeed !
The introduction of TryBegin was quite a challenge at the time and I never took the time to think about the possible consequences pour the CFG. Currently, we only introduce block for opcode appearing on opcode.hasjmp. We could extend the logic to opcode in opcode.hasexp when inside a TryBegin (note however that any such opcode could cause the execution of a code block to terminate when encountered outside of a TryBegin). This would likely require to keep track of the target block in case an exception is raised.
I would be open in a PR in that sense if the splitting of block for exception was optional.

Note that the approach you describe is somewhat wrong since it is never the TryBegin that will cause the jump.

BergLucas:
Yes, you're right. When I said that the `TryBegin` instructions caused the jumps, I meant that it was in these instructions that we specified the basic blocks that should be executed in case of exceptions.

I was mainly making this comment because at first, I didn't notice that these instructions weren't necessarily at the end of the basic blocks as, in Python 3.10, their equivalent were the `SETUP_FINALLY` instructions which were in `opcode.hasjrel` so they were always at the end and I expected the `TryBegin` instructions to have the same behaviour.

MatthieuDartiailh:
Could you describe a bit more what is your use case and what invariants you need the CFG to uphold ? 
I think there is room for improvement, but I would like a concrete example to design the right API.

BergLucas:
Hi,

Thank you very much for your quick responses!

To summarise, my use case is to create a `networkx.DiGraph` that matches a `ControlFlowGraph` so that I can navigate it more easily and add some instrumentation instructions in the right places.

Here is an example of code with explanations below:

```python
from bytecode.instr import Instr, TryBegin, InstrArg
from bytecode import Bytecode, ControlFlowGraph
import matplotlib.pyplot as plt
import networkx as nx


def foo(x: int):
    try:
        return 10 / 0
    except ZeroDivisionError:
        print("Division by zero error")


cfg = ControlFlowGraph.from_bytecode(Bytecode.from_code(foo.__code__))


def display_instr(instr: InstrArg):
    if isinstance(instr, Instr):
        return f"{instr.name} {instr.arg if isinstance(instr.arg, (int, str)) else ''}"

    return str(instr.__class__.__name__)


def display_graph(graph: nx.DiGraph):
    pos = nx.nx_agraph.graphviz_layout(graph, prog="dot")
    labels = {
        n: "\n".join(display_instr(instr) for instr in d["basic_block"])
        for n, d in graph.nodes(data=True)
    }
    nx.draw(graph, pos, labels=labels)
    plt.show()


def convert3_10(control_flow_graph: ControlFlowGraph) -> nx.DiGraph:
    graph = nx.DiGraph()

    for basic_block in control_flow_graph:
        graph.add_node(id(basic_block), basic_block=basic_block)

    for basic_block in control_flow_graph:
        jump = basic_block.get_jump()
        next_block = basic_block.next_block

        if next_block is not None:
            graph.add_edge(id(basic_block), id(next_block))

        if jump is not None:
            graph.add_edge(id(basic_block), id(jump))

    return graph


display_graph(convert3_10(cfg))


def convert3_11(control_flow_graph: ControlFlowGraph) -> nx.DiGraph:
    graph = nx.DiGraph()

    for basic_block in control_flow_graph:
        graph.add_node(id(basic_block), basic_block=basic_block)

    for basic_block in control_flow_graph:
        last_instr = basic_block[-1]
        if isinstance(last_instr, TryBegin):
            jump = last_instr.target
        else:
            jump = basic_block.get_jump()

        next_block = basic_block.next_block

        if next_block is not None:
            graph.add_edge(id(basic_block), id(next_block))

        if jump is not None:
            graph.add_edge(id(basic_block), id(jump))

    return graph


display_graph(convert3_11(cfg))
```

## Python 3.10

The `convert3_10` function shows a simplified implementation of what I was doing in Python 3.10 to create a `DiGraph` matching a `ControlFlowGraph`. In this case, the invariant that exists on `BasicBlock`s stating that we can only jump to another `BasicBlock` at the last instruction of the block is respected.

The use of the `convert3_10` function on the `foo` function gives the following `DiGraph`:



In this graph, we can see that there is a `SETUP_FINALLY` instruction that indicates the `BasicBlock` to execute if an exception is raised.

## Python 3.11

In Python 3.11, the `SETUP_FINALLY` instruction was replaced by a new system that significantly changed the way exceptions are handled. You managed this by creating the `TryBegin` and `TryEnd` instructions, which, in my opinion, represent the new system in a rather elegant way. Although these instructions do not actually exist in the Python bytecode, I thought that from the viewpoint of the `bytecode` library, they would work as if they were real instructions and follow the same rules as real instructions. Therefore, I assumed that the invariant on `BasicBlock`s would be respected and that by looking only at the last instruction, I would be able to easily find the `BasicBlock` to which we jump when an exception is raised. Thus, I updated the `convert3_10` function, which resulted in the `convert3_11` function. However, currently, the `TryBegin` instructions may violate the invariant, which may be the expected behaviour, but it breaks the `convert3_11` function. So I wondered whether it was a bug or the expected behaviour that the `TryBegin` instructions were not necessarily at the end of the `BasicBlock`s, given that their equivalent in Python 3.10 (the `SETUP_FINALLY` instructions) were basically doing the same thing and were always at the end.

The use of the `convert3_11` function on the `foo` function gives the following broken `DiGraph`:



It was more of a thought that occurred to me than a change I absolutely need in the library. If the current behaviour is the one you think is the best, then leave it as it is, and I'll use the `split_block` function to enforce the behaviour I want only in my use case.

PR #172, opened 2025-10-17
perf: improve general performance

P403n1x87:
We try to use sets for quick lookups and cache the result of some repetitive operations to speed up some operations. We also remove a good deal of asserts embedded within the code that adde extra overhead. Where appropriate, these should be replaced with actual tests.

The current change squeezes about 3/4% performance from a decompile/recompile loop.

## Benchmarking

The following round-trip script has been used to benchmark the performance of the proposed changes

```python
from time import monotonic

from bytecode import Bytecode

n = []
for _ in range(30):
    start = monotonic()
    for _ in range(2000):
        Bytecode.from_code(Bytecode.from_code.__code__).to_code()
    n.append(monotonic() - start)


avg = sum(n) / len(n)
sdev = (sum(_**2 for _ in n) / len(n) - avg**2) ** 0.5
print(f"{avg:0.2f} +/- {sdev:0.2f}")
```
Baseline: 1.10 +/- 0.02
This PR: 1.03 +/- 0.02

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
           pip install dist/*.tar.gz
           python -X dev -m pytest tests
       - name: Store artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: cibw-sdist
           path: dist/*
@@ -63,7 +63,7 @@ jobs:
           pip install dist/*.whl
           python -X dev -m pytest tests
       - name: Store artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: cibw-wheel
           path: dist/*.whl
@@ -79,7 +79,7 @@ jobs:
       id-token: write
     steps:
       - name: Download all the dists
-        uses: actions/download-artifact@v5.0.0
+        uses: actions/download-artifact@v6.0.0
         with:
           pattern: cibw-*
           path: dist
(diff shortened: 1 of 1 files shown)


src/bytecode/instr.py
```diff
@@ -22,46 +23,46 @@
...
-        (_opcode.opmap["LOAD_GLOBAL"], _opcode.opmap["LOAD_ATTR"])
+        {_opcode.opmap["LOAD_GLOBAL"], _opcode.opmap["LOAD_ATTR"]}
         if PY312
-        else ((_opcode.opmap["LOAD_GLOBAL"],) if PY311 else ())
+        else ({_opcode.opmap["LOAD_GLOBAL"]} if PY311 else set())
     )
 )
 
-BITFLAG2_OPCODES = (_opcode.opmap["LOAD_SUPER_ATTR"],) if PY312 else ()
+BITFLAG2_OPCODES = {_opcode.opmap["LOAD_SUPER_ATTR"]} if PY312 else set()
```
MatthieuDartiailh: Out of curiosity do you have a micro-benchmark for this. It is not obvious to me that for a single element a set is faster ?

P403n1x87: In fact I've got a pretty decent improvement when turning `HASJREL` etc... into sets. I'll post some benchmarking notes in the description.

P403n1x87: Just throwing an idea there: when the library is imported it is "locked" to the version of Python it is running under. There is no reason for a particular version of this library to support a range of Python interpreters. The only downside that I can see from migrating to a scenario where a release supports one specific Python release is that it would require some awkward dependency specification in other projects. However, in a [project that I work on](https://github.com/DataDog/dd-trace-py/blob/348e549220114e0daf15f3592579eed92d3746cd/pyproject.toml#L38-L42) this is already the case anyway. For example

```
bytecode~=0.17.0; python_version~=3.14.0
bytecode~=0.16.0; python_version~=3.13.0
...
```

MatthieuDartiailh:
Regarding version checks I wonder if we could try to cythonize the code and use tempita to actually remove the runtime version checks. Is this something you would be interested in ? H5py uses tempita with their cython code https://github.com/h5py/h5py/pull/2637

PR #176, opened 2025-11-21
feat: cythonise all sources

P403n1x87:
## Summary

Compile all bytecode sources as Cython extension modules to improve runtime performance. With plain Cython compilation (no type annotations), the speedup over pure Python was only ~3% (with 3.14, much higher with older versions), because every attribute access on the hot `BaseInstr`/`InstrLocation` objects still went through Python's generic attribute lookup (`_PyObject_GenericGetAttrWithDict`).

This PR adds `.pxd` declaration files that tell Cython these classes are `cdef class` extension types with C-level struct fields, eliminating the attribute lookup overhead for the hot paths.

### Key changes

- **`src/bytecode/instr.pxd`** — declares `InstrLocation`, `BaseInstr`, `Instr` as `cdef class` with `public` C-level attributes
- **`src/bytecode/concrete.pxd`** — declares `ConcreteInstr(BaseInstr)` with `_extended_args` and `_size` as `cdef public` fields
- `@cython.cclass` added to `InstrLocation`, `BaseInstr`, `Instr`, `ConcreteInstr` in their `.py` files (no-op when Cython is not installed, so pure Python mode is unaffected)
- `BaseInstr` drops `Generic[A]` base (incompatible with `cdef class`); adds `__class_getitem__` so `BaseInstr[int]` syntax still works in annotations
- `object.__new__` replaced with `cls.__new__` throughout fast-path constructors (`copy`, `_from_trusted`, `_from_opcode`, `_from_tuple`)
- `InstrLocation.__init__` and `_from_tuple` branch on `cython.compiled` to use direct assignment in compiled mode vs `object.__setattr__` in pure Python (where `@dataclass(frozen=True)` is still in effect)
- `.pxd` files are included in Cython wheel builds only (`package_data`)
- `setup.py`: `cython` added as unconditional `setup_requires`; `annotation_typing` left `False` to avoid treating function parameter annotations as C types
- `pyproject.toml`: `mypy` ignores type errors in the affected modules (dropping `Generic[A]` from `BaseInstr` cascades type errors on this branch that don't affect runtime correctness)

### Native CPU profile comparison

| Hotspot | Before | After |
|---|---|---|
| _PyObject_GenericGetAttrWithDict own | 5.47% | 4.28% |
| _PyObject_GenericGetAttrWithDict total | 19.74% | 13.20% |
| PyMember_GetOne (slot access) | 1.77% | eliminated |

PyMember_GetOne disappears entirely — slot-based member access is replaced by direct C struct field access. The remaining _PyObject_GenericGetAttrWithDict total comes from other classes not yet declared as cdef class (e.g. ConcreteBytecode, Bytecode, ControlFlowGraph).

### Throughput (Bytecode.from_code().to_code() round-trip, Python 3.14.4)

| Build | r/s range | median |
|---|---|---|
| Pure Python 3.14 | 125-130 | ~129 |
| Cythonized 3.14 | 153-163 | ~161 |

The Cython speedup over pure Python is ~25%.


P403n1x87:
@MatthieuDartiailh sorry for taking this long, this should now be good for review 🤞

MatthieuDartiailh:
Nice ! Do you have an estimate of the speed up ? 
Also is there any way to distinguish between the cython compiled wheel and the pure python wheels when installing ?

P403n1x87:
> Do you have an estimate of the speed up ?

I have added some benchmarks. CI might be too noisy to get reliable numbers, so probably best running locally.

> Also is there any way to distinguish between the cython compiled wheel and the pure python wheels when installing ?

I have added some output to the `setup.py`, but this won't be visible in general with `pip`. For tests, I have added a `conftest.py` to report the build type as part of the pytest headers. There are also test runs for both the cythonised and pure-Python versions.

MatthieuDartiailh:
I will merge after getting a chance to run benchmarks locally.

P403n1x87:
@MatthieuDartiailh i think this is as good as it gets for now. To push things a bit further we would have to refactor the code to avoid multiple inheritance, since that prevents us from turning more classes into more efficient c structs with Cython

src/bytecode/instr.py
```diff
@@ -690,11 +715,15 @@ def copy(self) -> TryEnd:
...
 
-class BaseInstr(Generic[A]):
+@cython.cclass
+class BaseInstr:
     """Abstract instruction."""
 
     __slots__ = ("_arg", "_location", "_name", "_opcode")
 
+    def __class_getitem__(cls, item: Any) -> Any:
+        return cls
```
MatthieuDartiailh: This should return a GenericAlias(cls, item)

.github/workflows/release.yml
MatthieuDartiailh, line 117: This will fail if cythonized and pure python wheels have the same name.

src/bytecode/instr.pxd
```diff
@@ -0,0 +1,14 @@
+cdef class InstrLocation:
+    cdef public object lineno
+    cdef public object end_lineno
+    cdef public object col_offset
+    cdef public object end_col_offset
```
MatthieuDartiailh: Why are those object rather than int ? Is it related to conversion cost ?

src/bytecode/instr.py
```diff
@@ -571,11 +584,17 @@ def __init__(
         col_offset: Optional[int],
         end_col_offset: Optional[int],
     ) -> None:
-        # Needed because we want the class to be frozen
-        object.__setattr__(self, "lineno", lineno)
-        object.__setattr__(self, "end_lineno", end_lineno)
-        object.__setattr__(self, "col_offset", col_offset)
-        object.__setattr__(self, "end_col_offset", end_col_offset)
+        if cython.compiled:
```
MatthieuDartiailh, line 603: Will cython respect the frozen nature of the class ?

src/bytecode/instr.py
```diff
@@ -690,11 +715,15 @@ def copy(self) -> TryEnd:
 A = TypeVar("A", bound=object)
 
 
-class BaseInstr(Generic[A]):
```
MatthieuDartiailh, line 693: If we cannot have a Generic because of Cython we should add .pyi files to preserve the type checking behavior.

.github/workflows/release.yml
P403n1x87: I believe they can't possibly have the same name because cythonised wheels will have at the very least a target CPython version tag, whereas the pure Python ones will have a generic tag.

src/bytecode/instr.py
P403n1x87: nope 🙁

src/bytecode/instr.pxd
P403n1x87: no it's because they can be `None` since they are all optional

src/bytecode/instr.pxd
```diff
@@ -0,0 +1,19 @@
+cdef class InstrLocation:
+    # Must be `object` (not `int`) because these fields are Optional[int] and can be None.
+    # `public` (not `readonly`) is required because _from_tuple/__init__ assign via an
+    # untyped `new` variable; Cython routes those through the Python descriptor, which
+    # would raise for `readonly`.  Immutability is enforced only in pure-Python mode by
```
MatthieuDartiailh: This really bothers me. Cannot we have a Cython private helper to be used for construction using direct access to mark those fields as readonly.

src/bytecode/instr.py
```diff
@@ -890,14 +923,6 @@ def __eq__(self, other: Any) -> bool:
...
 
-    _name: str
-
-    _location: Optional[InstrLocation]
-
-    _opcode: int
-
-    _arg: A
-
     def _set(self, name: str, arg: A) -> None:
```
MatthieuDartiailh, line 936: This typing looks wrong now.

src/bytecode/instr.pyi
```diff
@@ -0,0 +1,225 @@
...
+        location: Optional[InstrLocation],
+    ) -> T: ...
+    def has_jump(self) -> bool: ...
+    def is_cond_jump(self) -> bool: ...
+    def is_uncond_jump(self) -> bool: ...
+    def is_abs_jump(self) -> bool: ...
+    def is_forward_rel_jump(self) -> bool: ...
+    def is_backward_rel_jump(self) -> bool: ...
+    def is_final(self) -> bool: ...
+
```
MatthieuDartiailh, line 226: You need to add `_set` I think

MatthieuDartiailh (commented):
We are getting close. I will have to check manually what wheel pip will choose when confronted with both universal and cython compiled wheels.

src/bytecode/instr.pxd
P403n1x87: I've given it another go

setup.py
```diff
@@ -0,0 +1,50 @@
...
+        for p in get_py_files(ROOT)
+    ]
+
+
+_pure_python = os.getenv("BYTECODE_PURE_PYTHON")
+print(f"bytecode: building {'pure-Python' if _pure_python else 'Cython'} version")
+
+# Include .pxd declaration files only in Cython builds so they are available
+# to downstream Cython users who want to cimport from bytecode.
+_package_data = {} if _pure_python else {"bytecode": ["*.pxd"]}
```
MatthieuDartiailh: I have not dealt with that in a long time but since you specify files manually we may need to add .pyi now.

setup.py
```diff
@@ -0,0 +1,50 @@
...
+
+setup(
+    name="bytecode",
+    setup_requires=["setuptools_scm[toml]>=4", "cython"] + ([] if _pure_python else ["cmake>=3.24.2,<3.28"]),
```
MatthieuDartiailh: Why is cmake required ?

MatthieuDartiailh (commented):
Some last questions but we are getting there.

setup.py
P403n1x87: not required, fixed

MatthieuDartiailh:
I will get around to actually test and merge this great work but since a new release appears desirable in the short term (https://github.com/MatthieuDartiailh/bytecode/pull/179) I will proceed with is first and then come back to this and 3.15 support.

P403n1x87:
No rush on this, the other perf improvements seem to have been way more impactful than this anyway 😅

PR #191, opened 2026-04-07
Replace string-based references with postponed evaluation (PEP 563)

KishanBGajera:
This PR replaces string literal type annotations with postponed evaluation ([PEP 563](https://peps.python.org/pep-0563/)) for cleaner and more modern typing using:
```py
from __future__ import annotations
```

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -5,6 +5,8 @@ unreleased: Version 0.18.0
 --------------------------
 
 - drop support for Python 3.9 and 3.10 PR #180
+- Replace string literal type annotations with postponed evaluation using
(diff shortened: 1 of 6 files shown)


KishanBGajera:
@MatthieuDartiailh what do we think about this? :eyes:

MatthieuDartiailh (commented):
I am not a fan of the behavior of __future__ annotations since it is not really the future anymore.  However I must say it does hemp in cleaning up the code.
Could you add a changelog entry and I will merge ?

KishanBGajera:
@MatthieuDartiailh I updated it :)

MatthieuDartiailh (approved):

PR #194, opened 2026-05-08
perf: replace dis.get_instructions with direct co_code parsing in from_code

P403n1x87:
`dis.get_instructions` performs two full passes over the bytecode:
- `_make_labels_map → findlabels → _unpack_opargs` (to build a jump-label map)
- `_get_instructions_bytes` (to iterate instructions with full metadata)

Neither pass is needed here. `ConcreteBytecode.from_code` only needs the `opname`, raw arg byte, and source positions for each instruction word — all of which are directly available from `co_code` and `co_positions()`.

`CACHE` entries are already inline in `co_code` on all supported Python versions, so direct 2-byte iteration handles them naturally without the per-version `cache_info` loop that 3.13 previously required.

Throughput (round-trips of `Bytecode.from_code().to_code()` on the `dis` module's own code object, timed over 1 second, 3 runs each):

  Before: 92–94 round-trips/s
  After:  107–111 round-trips/s  (~+17%)

Own CPU time figures:

| Function | Before | After |
| --- | --- | --- |
| `dis._unpack_opargs` | 5.98% | 0.0% |
| `dis._get_instructions_bytes` | 3.45% | 0.0% |
| `ConcreteBytecode.from_code` | 3.63% | 4.91% |

### DIFF
--- a/src/bytecode/concrete.py
+++ b/src/bytecode/concrete.py
@@ -343,19 +343,15 @@ class ConcreteBytecode(_bytecode._BaseBytecodeList[Union[ConcreteInstr, SetLinen
         # available from Python 3.11+. CACHE entries are already inline in
         # co_code on all supported versions, so iterating co_code directly
         # handles all versions without dis overhead.
-        pos_iter: Optional[
-            Iterator[Tuple[Optional[int], Optional[int], Optional[int], Optional[int]]]
-        ] = iter(code.co_positions()) if hasattr(code, "co_positions") else None
+        pos_iter: Iterator[
+            Tuple[Optional[int], Optional[int], Optional[int], Optional[int]]
+        ] = iter(code.co_positions())
         for offset in range(0, len(bc), 2):
-            op = bc[offset]
-            arg = bc[offset + 1] if opcode_has_argument(op) else UNSET
-            if pos_iter is not None:
-                pos = next(pos_iter, None)
-                loc: Optional[InstrLocation] = (
-                    InstrLocation(*pos) if pos is not None else None
-                )
-            else:
-                loc = None
+            arg = bc[offset + 1] if opcode_has_argument(op := bc[offset]) else UNSET
+            pos = next(pos_iter, None)
+            loc: Optional[InstrLocation] = (
+                InstrLocation(*pos) if pos is not None else None
+            )
             instructions.append(ConcreteInstr(opname[op], arg, location=loc))
 
         bytecode = ConcreteBytecode()
--- a/src/bytecode/concrete.py
+++ b/src/bytecode/concrete.py
@@ -350,7 +350,7 @@ class ConcreteBytecode(_bytecode._BaseBytecodeList[Union[ConcreteInstr, SetLinen
             arg = bc[offset + 1] if opcode_has_argument(op := bc[offset]) else UNSET
             pos = next(pos_iter, None)
             loc: Optional[InstrLocation] = (
-                InstrLocation(*pos) if pos is not None else None
+                InstrLocation._from_tuple(*pos) if pos is not None else None
             )
             instructions.append(ConcreteInstr(opname[op], arg, location=loc))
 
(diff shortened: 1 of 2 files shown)


src/bytecode/concrete.py
```diff
@@ -336,21 +336,27 @@ def from_code(
...
+        # end_col_offset) per instruction word (including CACHE entries),
+        # available from Python 3.11+. CACHE entries are already inline in
+        # co_code on all supported versions, so iterating co_code directly
+        # handles all versions without dis overhead.
+        pos_iter: Optional[
+            Iterator[Tuple[Optional[int], Optional[int], Optional[int], Optional[int]]]
+        ] = iter(code.co_positions()) if hasattr(code, "co_positions") else None
+        for offset in range(0, len(bc), 2):
+            op = bc[offset]
+            arg = bc[offset + 1] if opcode_has_argument(bc[offset]) else UNSET
```
MatthieuDartiailh:
```suggestion
            arg = bc[offset + 1] if opcode_has_argument(op) else UNSET
```

src/bytecode/concrete.py
```diff
@@ -336,21 +336,27 @@ def from_code(
...
+        bc = code.co_code
+        opname = _opcode.opname
+        # co_positions() yields one (lineno, end_lineno, col_offset,
+        # end_col_offset) per instruction word (including CACHE entries),
+        # available from Python 3.11+. CACHE entries are already inline in
+        # co_code on all supported versions, so iterating co_code directly
+        # handles all versions without dis overhead.
+        pos_iter: Optional[
+            Iterator[Tuple[Optional[int], Optional[int], Optional[int], Optional[int]]]
+        ] = iter(code.co_positions()) if hasattr(code, "co_positions") else None
```
MatthieuDartiailh: Under what conditions could co_positions be missing ?

MatthieuDartiailh (commented):
Very nice !!! Getting rid of dis without complications is something I wished I had known was possible.

A couple of comments but LGTM

src/bytecode/concrete.py
P403n1x87: It was added in 3.11, so it might be safe to assume it won't be removed in future versions. In that case we can call it directly

MatthieuDartiailh: In that case a hard error would be better than silently missing information.

P403n1x87:
Throughput up to ~130 after updating the PR

src/bytecode/concrete.py
```diff
@@ -336,21 +336,23 @@ def from_code(
...
+        # co_positions() yields one (lineno, end_lineno, col_offset,
+        # end_col_offset) per instruction word (including CACHE entries),
+        # available from Python 3.11+. CACHE entries are already inline in
+        # co_code on all supported versions, so iterating co_code directly
+        # handles all versions without dis overhead.
+        pos_iter: Iterator[
+            Tuple[Optional[int], Optional[int], Optional[int], Optional[int]]
+        ] = iter(code.co_positions())
+        for offset in range(0, len(bc), 2):
+            arg = bc[offset + 1] if opcode_has_argument(op := bc[offset]) else UNSET
```
MatthieuDartiailh: Won't this be problematic for Cython ?

src/bytecode/concrete.py
```diff
@@ -336,21 +336,23 @@ def from_code(
...
+            pos = next(pos_iter, None)
+            loc: Optional[InstrLocation] = (
+                InstrLocation(*pos) if pos is not None else None
```
MatthieuDartiailh: This can use the new from_tuple constructor

src/bytecode/concrete.py
P403n1x87: It shouldn't be, the issue is specific to certain ad-hoc optimisations (cf. https://github.com/cython/cython/pull/7670)

MatthieuDartiailh: Actually in this particular occurance I do not find the walrus very readable. Could you go back to first fetching the op and then using it to get the arg ?

MatthieuDartiailh (approved):

PR #195, opened 2026-05-08
ci: add benchmark workflow

P403n1x87:
We add a custom benchmark workflow with an initial simple througput scenario that can give a quick feedback about the performance impact of each PR, compared to the base branch.


MatthieuDartiailh:
I feel like using something like codspeed would be easier and could provide more stable results (even though the raw number are not interpretable).

P403n1x87:
This is inspired by https://github.com/P403n1x87/austin/pull/402#issuecomment-4350906929 where I get a good indication about performance (modulo occasional noise, that is easy to spot anyway)

PR #198, opened 2026-05-10
perf: move BasicBlock type check from __iter__ to insertion methods

P403n1x87:
The isinstance check on every yielded item in BasicBlock.__iter__ was purely defensive — it detected invalid types that had already entered the list. Moving it to append/extend/insert/__setitem__ catches invalid types at the point of insertion instead, eliminating the check from the hot iteration path entirely.

The structural checks (jump must be last, jump/TryBegin target must be a BasicBlock) remain in __iter__ as they depend on the full block context and cannot be verified at insertion time.

Profiling data

| Hotspot | Before | After |
|---|---|---|
| `BasicBlock.__iter__` own | 4.91% | 2.82% |
| `ControlFlowGraph.from_bytecode` own | 4.61% | 4.04% |

Throughput (Bytecode.from_code().to_code() on dis module's code object, 1 second timed window, 5 runs):

| | r/s range |
|---|---|
| Before | 133–144 |
| After | 142–144 |

### DIFF
--- a/tests/test_cfg.py
+++ b/tests/test_cfg.py
@@ -55,13 +55,14 @@ def disassemble(
 
 class BlockTests(unittest.TestCase):
     def test_iter_invalid_types(self):
-        # Labels are not allowed in basic blocks
+        # Labels are not allowed in basic blocks — caught at insertion time
         block = BasicBlock()
-        block.append(Label())
         with self.assertRaises(ValueError):
-            list(block)
+            block.append(Label())
         with self.assertRaises(ValueError):
-            block.legalize(1)
+            block.extend([Label()])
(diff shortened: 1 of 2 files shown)


tests/test_cfg.py
MatthieuDartiailh, line 1: A test for `__setitem__` is missing.

MatthieuDartiailh (commented):
I would like to see an extra test but otherwise LGTM

MatthieuDartiailh (dismissed):

MatthieuDartiailh:
Actually looking at codecov report you are missing test cases for valid setitem (index and slice).

MatthieuDartiailh (approved):

PR #199, opened 2026-05-18
perf: replace O(n) name/varname list scan with O(1) dict lookup

P403n1x87:
_ConvertBytecodeToConcrete.add() used list.index() to find whether a name or varname was already registered, making each lookup O(n) in the number of names accumulated so far. For functions with many locals or names this becomes a non-trivial cost on every HAS_LOCAL / HAS_NAME instruction.

Replace the generic static add() method with two instance methods, add_varname() and add_name(), each backed by a parallel dict (varnames_map / names_map). Lookups are now O(1). The argnames pre-population loop is also updated to go through add_varname() so the map stays in sync.

Benchmark (round-trip test):
| | r/s |
| --- | --- |
| Baseline | 204.1 +/- 4.2 |
| This change | 221.3 +/- 2.9 |

So a +8.4% improvement overall. We also move some more instruction validation checks from iteration to insertion, which also contributes to the total uplift (~ +3% of the +8.4% total).

### DIFF
--- a/tests/test_cfg.py
+++ b/tests/test_cfg.py
@@ -82,19 +82,16 @@ class BlockTests(unittest.TestCase):
         block[:] = [nop]
         self.assertEqual(len(block), 1)
 
-        # Only one jump allowed and only at the end
+        # Only one jump allowed and only at the end — caught at extend time
         block = BasicBlock()
         block2 = BasicBlock()
-        block.extend(
-            [
-                Instr("JUMP_FORWARD", block2),
-                Instr("NOP"),
-            ]
-        )
         with self.assertRaises(ValueError):
-            list(block)
-        with self.assertRaises(ValueError):
-            block.legalize(1)
+            block.extend(
+                [
+                    Instr("JUMP_FORWARD", block2),
+                    Instr("NOP"),
+                ]
+            )
 
         # jump target must be a BasicBlock
         block = BasicBlock()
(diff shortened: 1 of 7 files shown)


MatthieuDartiailh (commented):
This makes sense I think even though it is a breaking change for CFG. Could you add a changelog entry so that I do not forget to document when doing the next release ?

P403n1x87:
> it is a breaking change for CFG.

In what sense? The only change in CFG is the time when validation happens, but no API has been changed (`_ConvertBytecodeToConcrete` is to be considered private given the `_` prefix I guess). Is this what you would like to have documented?

MatthieuDartiailh:
The fact the validation of the content now occurs earlier could break some usages (extend node and then break it into sub nodes). It is not a large breakage but worth documenting.

MatthieuDartiailh (dismissed):

MatthieuDartiailh:
Looking at codecov report we are actually missing tests for the error paths of append and insert. Would you mind adding those ?

tests/test_cfg.py
MatthieuDartiailh, line 116: Can you also cover the case of a TryBegin not targeting a block ?

MatthieuDartiailh: Can youvrename this test so that it is more descriptive (test_inserting_invalid_types) ?

MatthieuDartiailh (commented):
I promise those are the last comments. Recently I only manage to review from my phone which makes me miss the broader context quite often sorry for this.

P403n1x87:
> I promise those are the last comments.

No worries! Changes made.

MatthieuDartiailh (approved):

PR #205, opened 2026-06-03
fix: match correct try block markers in CFG

P403n1x87:
Fix an ``AssertionError`` in stack-size computation when a single exception region is split into multiple ``TryBegin`` instances sharing one handler (as produced by bytecode-rewriting tools that wrap a whole function body in a single handler). ``TryBegin``/``TryEnd`` are now matched on the handler block rather than on ``TryBegin`` identity.


P403n1x87:
@MatthieuDartiailh I think this is late to the 0.18.0 party 🙁 Could we have a 0.18.1 once this is merged? 🙏

doc/changelog.rst
```diff
@@ -1,6 +1,17 @@
 ChangeLog
 =========
 
+Unreleased
```
MatthieuDartiailh:
```suggestion
Unreleased: 0.18.1
```

MatthieuDartiailh (commented):
LGTM

Let me know when this is ready and I will release ASAP

P403n1x87:
@MatthieuDartiailh this should now be ready 🙇

MatthieuDartiailh (approved):

